PR #34, opened 2017-08-17
Fix people home page links

justinccdev:


AlasdairGray:
Thanks for spotting this bug and fixing it :)

PR #35, opened 2017-09-05
Updated Beacon affiliation to ELIXIR/GA4GH

susheel:


AlasdairGray:
This will now need to be done through the mapping document which is used to generate the specification display.

PR #38, opened 2017-10-05
Fix small typo in new spec page

aurel-l:


guicalman:
Thank you, hadn't notice that mistake.

PR #60, opened 2018-03-25
Feature/table layout update

ricardoaat:
Formatting, Adding Draft Gene profile, creating specification table template, using it on draft beacon spec. Enable SASS for the specification table template.

### DIFF
--- a/_devSpecs/Beacon/specification.html
+++ b/_devSpecs/Beacon/specification.html
@@ -22,10 +22,154 @@ subtitle: 'A convention for beacon to self-describe. '
 url: Beacons
 use_cases_url: /useCases/Beacons/
 version: '0.2'
+spec_info:
+  title: Beacon
+  subtitle: 'A convention for beacon to self-describe. '
+  description: 'In this document we propose a simple way for a beacons to self-describe
+    their genetic variant cardinality service for better integration with other beacons
+    within the beacon-network. It builds upon the Beacon service API and uses existing
+    schema.org entities and properties. '
+  version: "0.2"
+  official_type: ""
+  full_example: ""
+mapping:
+- property: description
+  expected_types:
+  - Text
+  description: A description of the item.
+  type: ""
+  type_url: ""
+  bsc_description: Description of this Beacon
+  marginality: Recommended
+  cardinality: ONE
+  controlled_vocab: ""
+  example: ""
+- property: identifier
+  expected_types:
+  - PropertyValue
+  - Text
+  - URL
+  description: The identifier property represents any kind of identifier for any kind
+    of Thing, such as ISBNs, GTIN codes, UUIDs etc. Schema.org provides dedicated
+    properties for representing many of these, either as textual strings or as URL
+    (URI) links. See background notes for more details.
+  type: ""
+  type_url: ""
+  bsc_description: Identifier
+  marginality: Recommended
+  cardinality: ONE
+  controlled_vocab: ""
+  example: ""
+- property: name
+  expected_types:
+  - Text
+  description: The name of the item.
+  type: ""
+  type_url: ""
+  bsc_description: Name of this Beacon
+  marginality: Minimum
+  cardinality: ONE
+  controlled_vocab: ""
+  example: ""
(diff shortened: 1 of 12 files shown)


ljgarcia:
Looking good but full example is missing, I am about to add it. Please do not merge yet, I'll use this branch to add the example.

ljgarcia:
Done, [pull request on the specifications repo](https://github.com/BioSchemas/specifications/pull/152) Once the other one is merged into master, we can update the link to full examples in the spec.

ljgarcia:
btw, this pull request looks fine to me but I think it is better that someone from managing the github.io takes a look/accepts/merges this PR. @AlasdairGray @kcmcleod @rajido please take a look, I have also created a [PR to host the Gene profile spec at the specifications repo](https://github.com/BioSchemas/specifications/pull/152).

kcmcleod:
@ricardoaat - a good first effort. However, there a couple of issues:

1. people like the property list to be sorted alphabetically. at the moment you have name, potentiationAction, url, dataset, provider ...
2. i like the colouring of the Minimum, Recommended and Optional words; however, others have claimed it is distracting.
3. you seem to have dropped/missed the links from the schema.org descriptions? For example, http://schema.org/identifier has a link on 'background notes'. 
Likewise, do you have the ability to keep any links added in the BSc description?
4. the BSc descriptions need to have the wording/grammar improved. i doubt we will ever be able to simply pull these from a spreadsheet without first reading/correcting them.
5. what should the examples look like? on my screen it is a small grey blob, which looks very weird. we need something better when there is no example.

ljgarcia:
@kcmcleod could you please point to specific errors in the BS descriptions? I would say that is not something thar @ricardoaat has to correct but the one creating the mapping (myself for the Gene case). Could we also have a native speaker helping with it as editor?

ljgarcia:
@kcmcleod @ricardoaat In order to allow links embedded in the text for Bioschemas description or CV column. What if we use the markdown way? anything that looks like [text](url) will be interpreted as a URL by GOWeb.
I tried to find a way in GoogleSpreadsheets but did not find any. If such way exist, the we should just use that way. Suggestions?

ricardoaat:
@kcmcleod Hi Kenneth, thanks for the feedback.

1. The template takes the order of the information on the YAML, This could be arrange on the tool that generates the YAML. I'll work on that.
2. I will remove the color of the marginality agrupation tag.
3. Can't you provide more information? Is this happening on the draft specification of Gene and Beacon? Those were the ones i've modified. 
4. I agree, The idea is to correct them on the spreadsheet. And not on the YAML file. Keeping the process as simple as possible. 
5. Beacon doesn't have examples for parameters, but Gene does. You can see how will it be displayed there. I could hide the ones that doesn't have examples though. 

@ljgarcia Hi Leyla, Currently the fields description and CV are being treated as plain text. We could think on a different object for that field. Perhaps a csv format on the spreadsheet?

kcmcleod:
@ljgarcia  @ricardoaat -  GOWeb seems a good first step.

> I tried to find a way in GoogleSpreadsheets but did not find any. If such way exist, the we should just use that way. Suggestions?

I agree that it seems difficult to do. I don't think we want contributors to write html or markdown in the cells. One suggestion was that we turn all instances of property or profile/type names into a link. However, it may be difficult to recognise these? 

It is also notable that Schema.org are inconsistent in the way they include links in descriptions. For example, look at the description for http://schema.org/image. Notice that it includes links to URL and ImageObject. In contrast, http://schema.org/accountablePerson does not link to either Person or CreativeWork.

> could you please point to specific errors in the BS descriptions? 

In the case of Beacon they are all missing a full stop.
I completely agree that @ricardoaat cannot be expected to correct mistakes, but wonder if there is a simple spell/typo checker that could help fix simple mistakes such as these?

ricardoaat:
@kcmcleod @ljgarcia Hello!
I've added the modifications discussed, except for the URLs on the description fields. That could be part of another feature in which i'd start working right away. This way we can merge this one and open path to a new one i have ready to PR.

Thanks!! 😃

PR #63, opened 2018-03-29
Feature/website responsiveness improvements

ricardoaat:
Since we've added new fields to the specification tables, it was necessary to make the following changes: 
- Some UI/UX and responsive improvements. 
- Sticky navbar for small devices integrated with the hamburger button. Replaced image hamburger icon with font awesome one.
- Added overflow to tables that were breaking the main container size.
- Adjusting some CSS naming for convention.


ricardoaat:
@kcmcleod @ljgarcia 
Header background color change was deliberated indeed. I haven't work on the footer yet, but i adjusted the logo's background color nevertheless.
Regarding the example's cell, now it will be shown only if an example is provided; it's now displaying the brake lines and spaces provided on the mapping spreadsheet.
@ljgarcia I will create another branch, thus another pull request to address the links issues and others issues that may (and most likely will) come up with Gene's specifications. Also, can i have more about the double scrollbar? on which SO and web browser is this happening? 
I have updated the PR with this.
Thank you both for the prompt and rich feedback, it's very helpful!

PR #67, opened 2018-04-16
Removed candidates table

ricardoaat:
- Table and tab candidates from /liveDeploys page removed.
- Better placing for bioschemas button styles.


AlasdairGray:
Eventually we will need to separate out by profile type, or at least filter.

PR #68, opened 2018-04-18
Feature/updating live deploys

ricardoaat:


ricardoaat:
@kcmcleod branch updated with master, you shouldn't see that text on the homepage anymore.

PR #85, opened 2018-05-10
Added ECCB and Paris BioHackaton to the event list

ricardoaat:
Also added markdownify to venue fields.

### DIFF
--- a/_layouts/meeting.html
+++ b/_layouts/meeting.html
@@ -3,7 +3,7 @@ layout: default
 ---
 <h1>{{ page.name }}</h1>
 <p><strong>Dates:</strong> {{page.dates}}</p>
-<p><strong>Venue:</strong> {{page.venue}}</p>
+<p><strong>Venue:</strong> {{page.venue | markdownify }}</p>
 {% if page.meeting-url %}
   <p><strong>Meeting page:</strong> <a href="{{page.meeting-url}}">{{page.meeting-url}}</a></p>
 {% endif %}
--- a/_meetings/2018-06_ELIXIRAllHands2018.html
+++ b/_meetings/2018-06_ELIXIRAllHands2018.html
@@ -5,6 +5,6 @@ title: Bioschemas for Datasets
 dates: 7 June 2018
 start_date: 2018-06-04
 end_date: 2018-06-07
-venue: Hotel Berlin, Lützowplatz 17, Berlin, Germany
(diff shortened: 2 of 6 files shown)


ricardoaat:
@rajido I've added NETTAB event, the only thing missing would be the exact location. Which I can add later as soon it gets confirmed. Please give it a look and let me know if it need changes 😃

ricardoaat:
@ljgarcia @rajido  Added start and end date detail on the meeting template. Also a note for the NETTAB event. Should I remove the BioHackathon, I think it's very important to have it there and even if the workshop proposal is not accepted, Bioschemas will be an important topic on the BioHackathon.

ljgarcia:
We could have the BioHackathon there if we add something like "submission in preparation". @rajido what do you think?

AlasdairGray:
I think that the Meetings page should only be for events that we have confirmed.

ricardoaat:
Wouldn't this mean ECCB and NETTAB event should be removed as well? Since they haven't been confirmed either.

AlasdairGray:
> Wouldn't this mean ECCB and NETTAB event should be removed as well? Since they haven't been confirmed either.

Yes. We should not say that we are running an event until we have confirmation.

I don't see any of these listed on the website at the moment.

ricardoaat:
I have gathered further information from Rafa regarding the events; both of them ECCB and NETTAB are confirmed as well as the BioHackathon since it's being organised by us. 

I have added a "To be confirmed" message to be removed after the workshops get properly advertised.

ljgarcia:
Only the first two have been officially confirmed. The Paris 
BioHackathon call is still open. Yes, we are confident Bioschemas will 
be accepted but it has not been yet. I would be cautious there.

rajido:
I think it is important we let the community know we are planning to have an event even if it hasn't been accepted yet. If we have submitted or we are planning to submit a proposal I think we should show the event and show a TBC. What do you think?

ljgarcia:
+1 for submitted/TBC events but separated some how for those already confirmed... maybe smaller font or so

ricardoaat:
I could of add a "(TBC)" tag on the title as well. What do you think?

ricardoaat:
I've added the badge. Please give it a look and let me know what you think.

AlasdairGray:
I don't think that they should appear on the front page. Perhaps we should have another section on the meetings page for events that we are only in the preparation phase for.

http://www.macs.hw.ac.uk/~ajg33

PR #87, opened 2018-05-11
Added Elixir TeSS event to live deploy list

ricardoaat:


ricardoaat:
@kcmcleod @ljgarcia It is a good idea, I will work on this on another branch.

PR #89, opened 2018-05-14
adding tess sources that use BSc markup

kcmcleod:
Adding sources that use working BSc markup for event. Ignoring those that have broken markup or just use vanilla schema.org


ricardoaat:
Do you think it'd be a good idea to add the ignored ones into the candidates.md file? We could use that list along with the crawler occasionally to check if the markup is included.

PR #91, opened 2018-05-15
Added HmmerWeb to list of LiveDeploys

aurel-l:
Not sure if the url is OK.
Only results are annotated with BioSchemas, but result pages disappear after a week, so can't really use a specific result page URL here.


ricardoaat:
Closing since these changes were added on pull request https://github.com/BioSchemas/bioschemas.github.io/pull/97

Thank you Aurélien

PR #94, opened 2018-05-16
Live deploy sorted by name and grouped by Profile

ricardoaat:


ricardoaat:
Thank you for your feedback. I have implemented an Accordion-like functionality for the table. Please give it a look and tell me what you think.
@kcmcleod @ljgarcia

ljgarcia:
Should not the icon be - when the accordion is open? Also... any way to improve the change of the width? It feels odd closing the accordion and then have to move the mouse to open it again just because of the change of the width.

ricardoaat:
@ljgarcia @kcmcleod  I've made some changes regarding the table size and the icon collapse indicator.

ljgarcia:
Table still resizes but it is better now. Icons look good although there is some sort of delay for the change, maybe related to the table resize? If no objections, I would say we can go with this version so we get the grouping in place and see how to improve the resizing later. @kcmcleod @AlasdairGray @rajido what do you think?

PR #98, opened 2018-05-25
Updated live deploy status for current and draft specifications

ricardoaat:
Addressing https://github.com/BioSchemas/specifications/issues/174


ricardoaat:
@kcmcleod Its an excellent idea, I will work on it as a feature.

PR #103, opened 2018-06-14
Ric/feat/people meeting relation

ricardoaat:
Changes to the way the data is related to one and another:

- Groups will have a member list instead of people having a list of groups they are part of.
- Attendees list will be on the meeting object.
- Added attendees to some events. In case the attendee is not on the _people repository, its ID will be shown. 
- Fixed minor bug on live deploys Visualise field.

This address the issue https://github.com/BioSchemas/specifications/issues/163


AlasdairGray:
Looks good. 

Just be careful not to do bug fixes on a feature branch, i.e. your fix to the live deploys page. That doesn't seem to have anything to do with this feature.

ricardoaat:
@AlasdairGray Thank you! I completely agree with you. It is just that the message on Jekyll's log for that bug was very annoying.
So, If everything it's ok with you, with your approval I will be able to merge this PR.

PR #107, opened 2018-06-20
Publications

kcmcleod:
Before I ask Alasdair for feedback, what do you think of the publications page?


ricardoaat:
The label for the link on the About>Presentation perhaps should include "Publication" word.

kcmcleod:
@ricardoaat  Thanks for the feedback!

Just wondering... are presentations & publications distinct? Should they have a page each? Should the title of the page (and the menu entry) just be Publications?

My current thoughts: publications seems to be a more general term and thus covers presentations. The URL of the page is /publications. Therefore the menu entry should be 'Publications'. Agree?

AlasdairGray:
I've made various updates to the layout of the page.

AlasdairGray:
Not needed for the page to go live, but the content of issue [192](https://github.com/BioSchemas/specifications/issues/192) should be added to the page.

AlasdairGray:
Is there a reason that the embedded slides are not the below code which shows all the slideshows in a single embedded object?

`<iframe src="https://www.slideshare.net/bioschemas/slideshelf" width="615" height="470" frameborder="0" allowfullscreen webkitallowfullscreen mozallowfullscreen></iframe>`

This is what is deployed on the current publications page?

PR #110, opened 2018-06-25
Publications

kcmcleod:
Added the missing publications (from https://github.com/BioSchemas/specifications/issues/188) to the /publications page. 

Thoughts?


ljgarcia:
@kcmcleod @AlasdairGray Could you please resolve the conflicts with the current master? Once that is done, I will review the branch. Thanks

AlasdairGray:
The merge conflict is due to more publications being added. I'll resolve this.

AlasdairGray:
@ljgarcia the conflicts have now been resolved.

It would be good to get this page up for now as it is more complete than the previous version. 

However, i think in the longer term we would be better using a liquid template that processes a bibtex or RIS representation.

AlasdairGray:
> Is it possible to have an orcid for a community? If so, we should create one for Bioschemas community so publications can be automatically populated from there... just an idea.

No idea about an orcid for a community. However, it will not necessarily help us as the website is only built when there is a change in the GitHub repository. 

We could look at having all the publications in a bib or ris file that is then processed by liquid commands in the web page

PR #112, opened 2018-06-26
Adding a set of slides to the front page

AlasdairGray:


ljgarcia:
I'm sorry but will not be able to review this week.

ricardoaat:
Two approvals will suffice for the creator of the PR to merge it. Adding @rajido as reviewer as well.

rajido:
I like the changes however I would suggest to change the layout to avoid the problems mentioned by @kcmcleod. I would suggest to have two columns. Left to show what Bioschemas is, relationship with schema.org and types. Right to show the upcoming events, the latest presentation and maybe tweets where Bioschemas is mentioned. What do you think?

justinccdev:
My main issue is with the text content (e.g. in the bottom para, Bioschemas is not (now?) looking to push biological types to schema.org itself).  I think it would also be worth spelling out the specific advantages of Bioschemas since text like "This structured information then makes it easier to discover, collate and analyse distributed data." applies to practically every other structured data initiative.  But this is just the same context as before this PR, so can be discussed later or separately.

Otherwise, it looks good to me (I'm not at all a design person).  One thing I would caution about the suggestion from @rajido is that displaying tweets by searching for 'bioschemas' is the type of thing that trolls have hijacked, though here it seems pretty unlikely anyone would bother.

AlasdairGray:
@ricardoaat I like that design for the desktop site. What happens for the mobile site?

AlasdairGray:
@justinccdev I agree that the text needs a revamp. We should raise that as a separate issue.

ricardoaat:
@AlasdairGray Currently, I have two options in mind:

1. The right bar to be stacked on the bottom of the left one making one big scrollable column. Exactly how's working now, but the slides and the twitter feed arrange at the bottom of the content.
2. An scrollable bar that would display the upcoming events, slides and twitter feed.
landing bioschemas
Something like this.

The first option is way easier to implement and perhaps the cleanest.

AlasdairGray:
I would go with the first option.

ricardoaat:
I think we all agree, will work on those now.

ricardoaat:
I have uploaded the proposed changes. Please give them a look and tell me what you think.

ricardoaat:
I like "What is Bioschemas?" black text is the default for <h1> which perhaps its better to leave as it is so its consistent with the rest of the site. Adding as well "Latest presentation" title to the slide section.

PR #115, opened 2018-06-26
Rename tools page to software

justinccdev:
As per https://github.com/BioSchemas/specifications/issues/193


AlasdairGray:
Need to change the mobile nav bar as well

ricardoaat:
It looks great! The only thing I see is we are adding a new link to the main menu. Which might have adverse effects on the UI. Perhaps we could have "Meetings" under the "About" sub-menu or something like that? @AlasdairGray Any thoughts?

AlasdairGray:
I agree, but the about menu is becoming a bit of a collect all.

I've started a discussion on this on https://github.com/BioSchemas/specifications/issues/199

However, for now I'm happy for the Software page to be added to the top level menu

PR #116, opened 2018-06-27
Create Taxon speciication folder

frmichel:
Hi @ricardoaat, here is a PR to create the Taxon spec folder including a first example to serve as a use case.


ricardoaat:
This is perfect! Merging this now

frmichel:
@ricardoaat, by the way, do I have to issue a PR any time I want to modify an example? Or would it be possible to give me write access on the Taxon folder and subfolders? (I'm afraid Github does not manage fine-grained rights, does it?).

PR #124, opened 2018-08-06
Added link to bioschemas slack workspace invitation

ricardoaat:
- Added invitation link to Bioschemas Slack's workspace.
- Added icons for the mailing list, github and slack link items.


justinccdev:
@AlasdairGray My belief was that it was meant to be used for anything.  But it is true that the only current use is software development (#awesomebuzzbang channel [I didn't come up with the name :)]).  So I think it does belong on the contact page, but add the qualification "for discussion of Bioschemas-related software development" rather than have people come there for other questions only to be met by tumbleweed.

ricardoaat:
Even thought the Bioschemas workspace on slack its being used mostly for tools development, since Leyla and I are always connected there we will be happy to help people with questions different from development tools.

PR #125, opened 2018-08-10
Feat/deployment stories

AlasdairGray:
I've added a way to list success stories of Bioschemas to the live deploy page. These show as a column to the right of the table of live deploys and link to external pages that give more details of the deployment story (see screenshot).

What do you think of the way I have done this? What could be improved?

screen shot 2018-08-10 at 16 04 37

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -48,7 +48,9 @@ collections:
     permalink: /:collection/:path/
   useCases:
     output: true
-    permalink: /:collection/:path/     
+    permalink: /:collection/:path/
   liveDeploys:
     output: true
     permalink: /:collection/:path/
+  stories:
+    output: true
--- a/_sass/_live-deploys.scss
+++ b/_sass/_live-deploys.scss
@@ -153,4 +153,10 @@
         font-size: 14px;
     }
 
+    .deployment-story {
+        margin-top: 1em;
+        border-radius: 5px;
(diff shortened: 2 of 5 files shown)


nsjuty:
I think we could condense display (current/draft/ specifications list is a bit repetitive)? Are the deployment stories just listed down the right? I wonder if they should be somehow linked to the resources on the left?

ricardoaat:
It looks really good, I would only add a bit more of margin between Stories, or perhaps wrapping them in a material-like card. Something like:
``` css
border-radius: 5px;
box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2);
```

AlasdairGray:
> It looks good, the only question is about maitaining this information up to date, any ideas there?

@ljgarcia One of the tasks for the implementation study is creating and maintaining this list. Hopefully this is a low effort way of doing this.

AlasdairGray:
> It looks really good, I would only add a bit more of margin between Stories, or perhaps wrapping them in a material-like card. Something like:

``` css
border-radius: 5px;
box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2);
```

@ricardoaat is it possible that you can do this?

ricardoaat:
@AlasdairGray I have added styling so the stories are contained on a material-like card. 
screen shot 2018-08-24 at 10 56 15

PR #126, opened 2018-08-15
adding dutch sites to live deploys

kcmcleod:
addressing https://github.com/BioSchemas/specifications/issues/201


AlasdairGray:
This appears to be correct to me.

AlasdairGray:
@ricardoaat is there anything we should be looking for when approving a pull request on the live issues page?

PR #133, opened 2018-09-14
Lj/feat/simpler draft spec path

ljgarcia:
In order to be able to have a jsonld for draft specs, we need to get rid of the /specification/ at the end. In this way, draft specs would look like http://bioschemas.org/devSpecs/Beacon/ and the correspoding definition on jsonld would be http://bioschemas.org/devSpecs/Beacon.jsonld. This is the way schema.org works, see https://schema.org/CreativeWork and https://schema.org/CreativeWork.jsonld.

We still need to do some more work for the properties, i.e., another pull request


ljgarcia:
@justinccdev @AlasdairGray @rajido @kcmcleod If no objections, I will merge today, cheers!

AlasdairGray:
We should ensure that each of the new pages has the redirection declaration from the previous URL. We now have a lot of broken links that have been previously advertised.

The following code needs to be inserted to the YAML at the start of the documents
```
redirect_from: "specifications/Beacon/specification"
```

PR #135, opened 2018-09-24
Created Gemfile for configuring test instance

AlasdairGray:
In setting up new test instances I've had to create this file. There seems to be some expectation that this would be in the repository.

@ricardoaat does it make sense to include this file in the repository?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,2 @@
+source 'https://rubygems.org'
+gem 'github-pages', group: :jekyll_plugins


ljgarcia:
Hi @AlasdairGray what is the purpose of having this file in the repo?

AlasdairGray:
I was receiving errors when running an instance on my own machine saying that the file was required.

@njall @ricardoaat any thoughts on whether this file is necessary or not?

kcmcleod:
Finn @ UoM says including the file is appropriate, but using it causes me problems. 

When I merge the master branch into this one, and run *jekyll serve* it fails with the error
```
`check_for_activated_spec!': You have already activated liquid 4.0.1, but your Gemfile requires liquid 4.0.0.
```
The Gemfile makes no mention liquid or any version numbers - not sure what is causing this error.
To solve this error you need to run
```
bundle exec jekyll serve
```
At which point it works.

ljgarcia:

AlasdairGray:
I don't know. I'm confused as to whether we need to supply a gemfile or not.

ljgarcia:
It works for me without it, I have not locally created one. That said, it might also be that I already have some things globally installed.

AlasdairGray:
As no-one seems to have had a need for it let's remove this PR.

AlasdairGray:
I just upgraded the OS on my laptop and needed this Gemfile to get the site working.

> It works for me without it, I have not locally created one. That said, it might also be that I already have some things globally installed. I tried this version locally and a side effect is that I have to add .jekyll-cache/Jekyll/ to the .gitignore.
> 
> Also, I get an error
> 
> ```
> /Ruby/lib/ruby/gems/2.6.0/gems/bundler-2.2.15/lib/bundler/runtime.rb:302:in `check_for_activated_spec!': You have already activated i18n 1.8.7, but your Gemfile requires i18n 0.9.5. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
> ```
> 
> that I do not get if I remove the gemfile so... not sure we need it

Do you use `jekyll serve` or `bundle exec jekyll serve`. I think the latter is the safer form. It looks like you may have some other ruby app that uses some different gem configurations.

AlasdairGray:
@albangaignard do you know what is needed in terms of Gemfiles?

albangaignard:
I'm running MacOS 11.4 without any Gemfile in my working copy, and it's working fine (`jekyll serve`). I would go without pushing it to the git repository.

AlasdairGray:
Strange, I've just upgraded to 11.4 and needed the Gemfile. It also seems to be required based on the GitHub pages [documentation](https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/creating-a-github-pages-site-with-jekyll#creating-your-site) (see step 8 onwards).

However, happy to add the contents of the Gemfile to the README if we feel that is a better approach.

PR #142, opened 2018-10-17
Completed rename of training group 

njall:
And added some GOBLET members to training group (more to come..)


njall:
Hi, I've started a bunch of commits on this fork - so when this is all groovy and can be merged I'll start working on branches of the main repo. 

So addressing the issues you picked out before (thanks for that!!)

- It seems that some former members are not listed, e.g. Phil Barker.
 
 - Done, added all the people currently engaged in developing specs to the training group. 

  - It was bugging me making YAML files so I wrote a script. Simply add your group members names and which group it is, and run to generate new YAML files for each  https://gist.github.com/njall/2221b572489a7270d2a8aaf766927d3b

- links to the specifications are broken

  - This is because they're in draft status. They can be found on devSpecs/<schemaname> but this automatically links to specs. Might need to add a separate field for in_dev_specifications. I've removed the links for now. We intend to release these specs in the new year, so I'll uncomment them then

- the list of contributors in a specification are missing.

  - Could you elaborate on this? Is this solved by my latest commits?

PR #148, opened 2018-11-10
Pruning out unused folders

AlasdairGray:
These folders were created by map2model but unused. Examples are all stored in the Specification repository


AlasdairGray:
@ricardoaat can you confirm that we do not need these folders?

PR #158, opened 2018-11-17
Ric/livedeploy/adding chembl implementation

ricardoaat:
ChEMBL now implements Dataset and MolecularEntity. Added those to the LiveDeploy page.


ricardoaat:
We do have plans to include DataCatalog on the website, but right now we have implemented DataSet and MolecularEntity on ChEMBL website, so for now I think it should be include on LiveDeploys Dataset and MolecularEntity sections only.

AlasdairGray:
The markup I looked at included DataCatalog markup

ricardoaat:
Maybe I've missed it, where can I find the markup you looked with DataCatalog?

AlasdairGray:
It was stated that the dataset was in a days catalog and then a description of the catalog was given. Did I misread something?

http://www.macs.hw.ac.uk/~ajg33

ricardoaat:
Yes, but it's far from complete. Since we are not including much information of the DataCatalog (not even the minimum for Bioschemas) just yet in the property "includedInDataCatalog", I don't think we could say we properly marked up the DataCatalog type on that JSON-LD. As soon as I have it online it, I believe we could have it as a live deploy.

AlasdairGray:
OK, that's a fair point. I only glanced through it.

ricardoaat:
Thank you!

PR #162, opened 2018-12-12
added alliancegenome  to live deploy

nathandunn:


nathandunn:
Our changes for Bioschemas / Gene are in there as well, but won't be in production until March.

AlasdairGray:
When you add them in we can update the live deploy page again.

PR #164, opened 2019-01-15
Data catalog v0.2 draft

AlasdairGray:
Draft specification updated based on discussions at Biohackathon. Changes here reflect the crosswalk spreadsheet.

**Change Log**

Property | Change | GitHub Issue(s) | Note 
-- | -- | -- | -- 
  | Other | #175 | Examples added to many properties 
keywords | Cardinality |   | Changed to many to allow multiple forms
rdf:type | Other | #146 | Removed from explicit set of properties, but still minimal; Now implicit 
about | Added |   | Enables link to the topic of the catalog 
identifier | Cardinality |   | Changed to many to allow multiple forms 
publication | Deprecated | #128 | publication was not being used correctly 
fileFormat | Deprecated |   | SupersededBy encodingFormat on schema.org 
encodingFormat | Added |   | Supersedes fileFormat on schema.org


ljgarcia:
hi @AlasdairGray 

Keywords, as defined in schema.org should be only one with commas separating multiple values. About identifier to MANY, not sure about it. Identifier should be used for the official one that will be mostly used for findability, any other could go to sameAs

AlasdairGray:
I guess this is down to interpretation of MANY. I'm easy either way as they would be semantically equivalent.

AlasdairGray:
@sarala has added examples for the minimal properties where these were missing. This PR also closes issue [128](https://github.com/BioSchemas/specifications/issues/128).

PR #167, opened 2019-02-04
Create people file for Yvan Le Bras

frmichel:
Yvan just joined the community (email exchanged with Leyla and I).

### DIFF
--- a/_people/ylebras.md
+++ b/_people/ylebras.md
@@ -1,0 +1,11 @@
+---
+layout: person-details
+id: ylebras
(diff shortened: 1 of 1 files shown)


frmichel:
Hi @ljgarcia, 

This means that each _people file points to the groups, and conversely that each _group file points to the group members.
Right? Not a problem to have this redundant information?

Franck.

ljgarcia:
@frmichel regarding your question, maybe @AlasdairGray can tell you better

PR #169, opened 2019-02-04
Biotea2bioschemas

ljgarcia:
Scholarly publication related articles started during the Paris BioHacakthon following the model proposed by Biotea


ljgarcia:
Pages working, specs are drafts so this can go in.

AlasdairGray:
It's unclear to me why we need such fine grained modelling of scholarly publications. Why do we need to distinguish between Journal and ScholarlyPublication.

Also, some of the links in the pull request do not work.SemanticScholarlyAnnotation (whatever that is) takes you to the Journal profile.

PR #179, opened 2019-04-04
Rephrased text and changed email address

AlasdairGray:


martin-nc:
Yes, that's better. Or you could make it even easier by adding the email address:

> ...let us know via the <a href="http://lists.w3.org/Archives/Public/public-bioschemas/">public mailing list <i class="fas fa-envelope"></i></a> (<a href="mailto:public-bioschemas@w3.org">public-bioschemas@w3.org</a>).

This way could go to the mailing list to see if a topic has already been asked, or get the email address without an extra click.

AlasdairGray:
I've now added that. If you are happy, please merge the pull request.

PR #180, opened 2019-04-08
Update software page

kcmcleod:
1. Moved BuzzBang into an Archive section at bottom of page.
2. Reordered the list of tools as GoCrawlIt seems like it should also be in Archive, so no longer 3rd item in list. 
3. Left Validata and MarkupBuilder as placeholders.

Other changes?


AlasdairGray:
I've made further changes:
- Moved request for updates to top of page
- Ordered tools based on 
  - Own tools over external (i.e. Google's generic search)
  - Tools that are in active development
- Minor corrections to layout and presentation

kcmcleod:
Fine.

Though could be made to look nicer as the spacing is off for the GoCrawlIt. Perhaps another job for Sean?

PR #181, opened 2019-04-12
Cached copy of Bioschemas proposal for schema.org

AlasdairGray:
For the 8 May meeting it would be good to have a copy of the proposal that 

1. Can be accessed by everyone without needing to install any software and 
1. Does not rely on the appspot site since this has a low daily usage limit

This pull request creates a new schema directory that contains a cache of the proposal, i.e. the full content of the appspot site is available from bioschemas.org/schema/

A potential issue with this approach is that people start editing the content here and this is then not reflected in the schemaorg proposal. However, we would have the git commits and would be able to see which files have been changed and then copy those changes by hand into the schemaorg proposal.

Another issue is that we already have the devTypes, e.g. [Protein](https://bioschemas.org/devTypes/Protein/). After the meeting, we will need to think about how best to proceed with our proposal versus the devTypes content.


kcmcleod:
You've cached the files in this repo, but I don't see an obvious link to the local copy. How do people know to access it? I assumed a link on the f2f meeting page, but I cant find one. 

> A potential issue with this approach is that people start editing the content here and this is then not reflected in the schemaorg proposal

As you say, I think that is simple (though perhaps) time consuming to resolve using the git commits.
Another potential problem is that if some work on the local copy and others the appspot copy, do you loose synergy when they are not (re)viewing each other's work?

ljgarcia:
+1 to @kcmcleod comments regarding editions on both places. 
And, I did not find a link to the schema.

AlasdairGray:
I was going to add the link to the F2F meeting once we got the cache copy on the website.

Only Leyla and I were editing the appspot version, so I don't think it is a huge concern of edits being in multiple places. A bigger concern is the low access limit we have at appspot which would prevent us from being able to view the content at all.

PR #182, opened 2019-04-25
Removing isa-explore from live deploys

kcmcleod:
http://scientificdata.isa-explorer.org/ is no longer online, ie, URL doesn't resolve


AlasdairGray:
The resource resolves for me and comes back with a DataCatalog when tested with the Google [tool](https://search.google.com/structured-data/testing-tool#url=http%3A%2F%2Fscientificdata.isa-explorer.org%2F)

kcmcleod:
Agreed, it is now back. Should delete PR

PR #187, opened 2019-05-08
changes as agreed from BioSchemas F2F

sillitoe:


AlasdairGray:
@sillitoe was a new version of the crosswalk spreadsheet created and used for these changes or is the only version of them in this pull request?

sillitoe:
Sorry, I'm not familiar with the protocol for releases / PRs

We used the following sheet:

https://docs.google.com/spreadsheets/d/1UbTy5OACa08ZTX9yk35AaVRUS6YBm1jBqwgG-Y41KGc/

Does that help?

AlasdairGray:
Thanks @sillitoe. I'll try to stitch things back together so that we have everything in the right place.

AlasdairGray:
I see that @ljgarcia created a new copy of the mapping file for you to create this version of the profile with. It all looks good to me.

PR #192, opened 2019-05-13
Remove outdated cache

AlasdairGray:
This PR removes the cached copy of the Bioschemas proposal for types and properties and replaces them with a placeholder index page that redirects to the latest version hosted on appspot. This is to eliminate the need to maintain the cached copy.

The PR also introduces a custom 404 page for when someone tries to visit a page that does not exist.


kcmcleod:
seems to work.
* http://localhost:4000/banana produces custom 404.
* http://localhost:4000/schema redirects to http://bio.sdo-bioschemas-227516.appspot.com/

PR #193, opened 2019-05-14
Added Gene use case

AlasdairGray:
This PR takes the Gene Use Case that was developed in [issue 247](https://github.com/BioSchemas/specifications/issues/247) and makes it available through the web site.

It also adds a profile for Chunlei Wu and adds him to the Gene group.

### DIFF
--- a/_groups/Genes.md
+++ b/_groups/Genes.md
@@ -45,4 +45,5 @@ members:
     - MichelDumontier
     - AndyJenkinson
     - RicardoArcila
+    - ChunleiWu
 ---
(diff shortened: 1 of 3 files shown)


kcmcleod:
Gene use cases available.

@AlasdairGray : 
* Profile for Chunlei Wu seems fine; have you asked him if he is happy with it?
* localhost:4000/useCase/Genes redirects to localhost:4000/useCase/Gene; as you said.
However, http://localhost:4000/specifications/Genes/ does not redirect nor does http://localhost:4000/devSpecs/Genes/. Should they?

AlasdairGray:
> * http://localhost:4000/specifications/Genes/ does not redirect nor does http://localhost:4000/devSpecs/Genes/. Should they?

I think these are a separate issue which we should address when we look at the Gene type.

PR #195, opened 2019-05-16
Update uc issue314

kcmcleod:
making use case names singular & adding plural redirects
updating links in specs & draft tables to singular use case pages

FIXES BioSchemas/specifications#314


AlasdairGray:
Use case file names still seem to be plural, e.g. useCase/Beacons.html. Am I missing something here?

PR #203, opened 2019-06-12
removing group leaders from members list

kcmcleod:
removes group leaders from members list of group pages as discussed in https://github.com/BioSchemas/specifications/issues/316


kcmcleod:
updated person-details to include groups that the person leads.

PR #205, opened 2019-06-15
Major cleanup of types and profiles

AlasdairGray:
The main reason for the major overhaul is to provide us with a clean story to those looking to engage, and those already engaged, with Bioschemas. See below for full details of the changes.

The main concern is that some groups who have been engaged with Bioschemas become disenfranchised since their type/profile have been demoted to the drafts pages. Our intention is it speed up the development on the remaining types and push these up to schema.org in the coming months and certainly before the BioHackathon.

Here is a summary of the changes:

**Types Page**
The types page would only contain those types that we are pushing towards schema.org in the first release. These have been updated to be the same as those on the appspot development site. 

**Draft Types Page**
The draft types page includes all the types that we have been working on. Those that do not have a June date still need to be updated with the version from the appspot development site, unfortunately this is a manual process.

I took the decision to revert LabProtocol to draft status as there are several issues that need to be resolved. It was the only type that did not have a newer draft version.

**Profiles Page**
The profiles page only contains those profiles that are over a stable schema.org type, one of the types that we are currently pushing into schema.org, or has been widely deployed (Sample). 

Profiles over the new types, e.g. ChemicalSubstance, Gene, and Protein, are not on there at this point since we need to either test these still with actual deployments/examples or update their profile description to the latest version of the type.

Based on my discussion with Niall, Event and TrainingMaterial have remained since they are used by TeSS. They will need revising in the near future, in particular Event since it includes properties that are not in schema.org or our proposal for schema.org. However, this is part of the push on Course and CourseInstance.

**Draft Profiles Page**
Includes the latest version of all profiles that are in development. There is a wide range of variance in the status of these from not currently being actively worked on, e.g. Standard, to almost ready for making into a stable profile, e.g. Course and CourseInstance. Several of them that relate to new types still need to be updated to reflect the new type and then need to be tested.


nsjuty:
looks good - sent comments via email re "v0.1-draft.1." nt being used, and no indication whether DRAFT vs draft has any significance

AlasdairGray:
> "v0.1-draft.1." nt being used

This has been removed in the clean-up. If you are still seeing this, can you indicate where it is.

> no indication whether DRAFT vs draft has any significance

No significance, just different people doing slightly different things. Personally I prefer 'DRAFT', but easy either way. I do agree that we should be consistent.

AlasdairGray:
Approval received from @CaroleGoble

PR #208, opened 2019-06-18
Fix the redirects on current types

AlasdairGray:
Found that some of the redirects were not correct. I have fixed these.


AlasdairGray:
I agree that the usage of DRAFT on the current page is wrong. Perhaps the term 'release candidate' is better.

The UX idea is nice as then we only need one page, but not sure that it is worth the effort.

AlasdairGray:
Chatted with @nsjuty at the coffee break just now and he agreed that 'release-candidate' (RC) would be better than the 'DRAFT' status on the current types page.

kcmcleod:
yes, good name!



--

Kenneth McLeod, PhD

Semantic Web Lab: http://www.macs.hw.ac.uk/SWeL/
School of Mathematical and Computer Sciences
Heriot-Watt University


[cid:7c5537cf-da2d-4ead-8b8a-eb906d542de2]

AlasdairGray:
I've made the change to each of the types and revised the text on the types page. Are we happy to merge these in now?

kcmcleod:
now the current types table has RC in the name.
the drafts types table has DRAFT in the name.
when you click on the name in either table you get to the RC version, which is still confusing.

perhaps better to remove the RC types from the draft page?

AlasdairGray:
Good catch. I think you are right that we should not include the RCs on the drafts page, it just causes confusion.

I've made it so that the drafts page of the RC types redirect to the RC version and removed the RC types from the drafts list. I also updated the text on the drafts webpage to indicate what is happening.

What is not very satisfying is that we do not have all the types that have been developed on the schemaorg system on the Bioschemas drafts type page, but this requires a manual effort and not something that we can do straightaway. For now I have listed those that I know to be missing for the purpose of keeping track of them and allowing others to know that we are at least working in those areas.

kcmcleod:
looks good

PR #209, opened 2019-06-18
Added Jerven to the people page and the Protein group

AlasdairGray:
@JervenBolleman are you happy with being added to the People list in Bioschemas?


JervenBolleman:
Yes, this is fine.

AlasdairGray:
Great. Are there any groups other than Protein that you would like to be listed on?

PR #210, opened 2019-06-20
Correct cardinality of dataset property on DataCatalog

AlasdairGray:
Fixes issue https://github.com/BioSchemas/specifications/issues/330
Also adds an example

@kcmcleod can you check that the example is correct

### DIFF
--- a/_devSpecs/DataCatalog.html
+++ b/_devSpecs/DataCatalog.html
@@ -5,7 +5,7 @@ redirect_from:
 - "/devSpecs/DataCatalog/"
 name: DataCatalog
 
-status: release #revision replace with revision for next version
+status: revision #revision replace with revision for next version
 spec_type: Profile
 group: datarepositories
 use_cases_url: /useCases/DataRepository/
@@ -25,8 +25,8 @@ spec_info:
   subtitle: ""
   description: A guide for how to describe data catalogs/repositories in the life-sciences
     using Schema.org-like annotation.
-  version: 0.2
-  version_date: 20190614T144341
+  version: 0.3-DRAFT
+  version_date: 20190620T095210
   official_type: http://schema.org/DataCatalog
   full_example: https://github.com/BioSchemas/specifications/tree/master/DataCatalog/examples/
 mapping:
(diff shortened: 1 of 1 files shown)


kcmcleod:
There is an update for the web page, but not the gSheet that the web page is derived from. Are we abandoning the gSheets and GoWeb?

WRT the example, what you have is valid markup; however, it is not applying the Bioschema's Dataset profile. Do we recommend people do this for nested entities or does it depend on whether or not it is defined elsewhere? If the later, is that documented somewhere?

AlasdairGray:
Adding comments here to summarise the conversation I've had with @kcmcleod. I believe that these issues have been fixed in commits https://github.com/BioSchemas/bioschemas.github.io/pull/210/commits/ba696aac834142b5f0d809de5f302d85c30c131c and https://github.com/BioSchemas/bioschemas.github.io/pull/210/commits/52e62aab19cc12e72dec2d07b65c7753c798bc8b. @kcmcleod please confirm.

> There is an update for the web page, but not the gSheet that the web page is derived from. Are we abandoning the gSheets and GoWeb?

No, I've probably forgotten to update one of the gSheet links. I created a new copy of the gSheet to correspond to this version of the profile.

We seem to have two properties that are maintaining a link to the gSheet, it would be good to eliminate this redundancy. I do not know which the web pages are actually using.

> 
> WRT the example, what you have is valid markup (I'll push a commit to fix the typo); however, it is not applying the Bioschema's Dataset profile. Do we recommend people do this for nested entities or does it depend on whether or not it is defined elsewhere? If the later, is that documented somewhere?

Ideally we would have two examples, one which shows how to link to another resource (which I believe would be the favoured approach), and one which shows the embedding of nested markup.

kcmcleod:
> No, I've probably forgotten to update one of the gSheet links. I created a new copy of the gSheet to correspond to this version of the profile.
Link has been changed

> We seem to have two properties that are maintaining a link to the gSheet, it would be good to eliminate this redundancy.
Removed for this page. However, other pages have this issue; these are not fixed yet (https://github.com/BioSchemas/specifications/issues/333)

PR #215, opened 2019-07-17
Live deploy add node cdr edd

kcmcleod:
Changes:
* Removed identifiers.org as it no longer has markup on homepage.
* Increased biosamples pages from 5 million to over 10
* added node info & CDR/EDD info
* changed stylesheet to ?improve? presentation of highlights & node/CDR info

@AlasdairGray not a fan of the way is this presented. What do you think?

### DIFF
--- a/_liveDeploys/liveDeploy.md
+++ b/_liveDeploys/liveDeploy.md
@@ -7,15 +7,6 @@ type: liveDeploy
 description: Services/sites implementing Bioschemas's markup
 list:
 -
-    name: Identifiers
-    highlight:
-    example_URL: http://identifiers.org/
-    resource_URL: http://identifiers.org/
-    schema_org: DataCatalog
-    bsc_profile: DataCatalog
-    bsc_ver: 0.1
-    comments: Could implement Dataset and DataRecord as well.
--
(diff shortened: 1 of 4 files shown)


AlasdairGray:
Thanks for doing this. At the moment it looks strange for non-ELIXIR resources which will not have a node associated with them or be eligible for CDR or DDR status. Therefore, I think it would be better for us to have a separate ELIXIR deployments page, populated from the same yaml file, that only has ELIXIR resources on it.

AlasdairGray:
> Removed identifiers.org as it no longer has markup on homepage.

I suggest that we only comment out Identifiers.org for now. They should have markup again soon.

kcmcleod:
@AlasdairGray  removed all the additions to the LD page. Added an elixir only page, which is linked from the LD page. Any better?

kcmcleod:
Hopefully tackled all critical tasks above PLUS both pages now are closed by default.

Adding ticket for desirable tasks.

AlasdairGray:
I've made some additional refinements to the page overview text and added the legend. 

@kcmcleod please verify and merge if correct.

PR #217, opened 2019-07-18
Quick fix tool

kcmcleod:
Simple changes for ismb demo:
* fixing typos on tool & correcting an example
* adding organizaton & person via the new gDoc


kcmcleod:
@AlasdairGray - didn't realise the HTML was hard coded at the bottom of ALL of these pages. Replaced with the standard template that uses the yaml to generate the page so they are now different to the pages you approved. 
In particular, Tool no longer has Person and Organization at the bottom of the page. Happy with this?

Corrected the version numbers & cross walk links too.

AlasdairGray:
The tools group had started their own versions of Person and Organisation, which is why they were hard-coded at the bottom.

However, it would be best if they contributed to the Person and Organization profiles given that they are not really being reviewed by anyone else.

PR #221, opened 2019-07-24
Fixed errors in properties

AlasdairGray:
The Taxon type has not yet been accepted into Schema.org so the properties should still be listed as bioschemas. I have updated the spreadsheet and the web site. I also found that the webpage had the link to the wrong spreadsheet so I updated that as well.


frmichel:
Hi @AlasdairGray, I'm ok with the changes. I've also updated the description in tab "Specification info" to reflect this. I worded it this way: "parentTaxon, childTaxon and taxonRank are in namespace bioschemas.org until they be eventually accepted into schema.org"

Is that ok with you?

AlasdairGray:
OK, that is a good call. I've made the webpage and the gSheet the same.

PR #225, opened 2019-08-20
Update protein profile

ljgarcia:


_devSpecs/Protein.html
```diff
@@ -60,20 +61,23 @@
   bsc_description: Disease associated to this protein, if any.
   marginality: Recommended
   cardinality: MANY
-  controlled_vocab: ""
+  controlled_vocab: Any suitable disease ontology
```
AlasdairGray: Should we list some, e.g. Disease Ontology, OMIM, Orphanet, etc

AlasdairGray: This is not a blocking issue for this PR. Now captured in issue https://github.com/BioSchemas/specifications/issues/355

PR #227, opened 2019-09-06
Removed escaping that caused Jekyll errors

egonw:


AlasdairGray:
Not seeing errors on my side, but happy to accept this change as it still seems to work with it.

PR #229, opened 2019-09-12
Group buttons

kcmcleod:
Adding buttons to the group pages that will email the leader(s) of the group to request membership
* **NOTE** this is not complete as I don't have a list of all emails
* Try Validation for a working example

Adding buttons that email Alasdair, Leyla & Carole to:
* groups page: propose a new group 
* profiles & draft profiles pages: propose a new profile
* types & draft types pages: propose a new type


groups/index.html
```diff
@@ -3,7 +3,13 @@
 title: BioSchemas Groups
 ---
 
-<h1>Bioschemas Groups</h1>
+<h1>Bioschemas Groups 
+<div class="git-edit" style="float: right;">
+<a target="_blank" href="mailto:leylajael@gmail.com,a.j.g.gray@hw.ac.uk,carolegoble@gmail.com?subject=[Bioschemas] request to create a new group" class="btn btn-bioschema btn-sm" role="button">
```
AlasdairGray: To avoid maintenance issues this should go to the enquiries@bioschemas.org email address, or we could even investigate setting up a mail alias for the management group.

groups/index.html
```diff
@@ -3,7 +3,13 @@
 title: BioSchemas Groups
 ---
 
-<h1>Bioschemas Groups</h1>
+<h1>Bioschemas Groups 
+<div class="git-edit" style="float: right;">
```
AlasdairGray: I wonder if this should be at the bottom of the page rather than the top so that it is obvious once you've scrolled through all the groups.

specifications/drafts.html
```diff
@@ -5,7 +5,11 @@
  - "/devSpecs/"
 ---
 
-<h1>Draft Bioschemas Profiles</h1>
+<h1>Draft Bioschemas Profiles
+<div class="git-edit" style="float: right;">
+<a target="_blank" href="mailto:leylajael@gmail.com,a.j.g.gray@hw.ac.uk,carolegoble@gmail.com?subject=[Bioschemas] request to create a new profile" class="btn btn-bioschema btn-sm" role="button">
```
AlasdairGray: Use a generic email address.

specifications/index.html
```diff
@@ -2,7 +2,11 @@
 layout: default
 title: BioSchemas Profiles
 ---
-<h1>Bioschemas Profiles</h1>
+<h1>Bioschemas Profiles
+<div class="git-edit" style="float: right;">
+<a target="_blank" href="mailto:leylajael@gmail.com,a.j.g.gray@hw.ac.uk,carolegoble@gmail.com?subject=[Bioschemas] request to create a new profile" class="btn btn-bioschema btn-sm" role="button">
```
AlasdairGray: Generic email address

AlasdairGray:
I wonder if we should only have the button on the draft type and profile pages.

groups/index.html
kcmcleod: Carole would say that no one scrolls to the bottom of the page; I know this as Alan said exactly that to me yesterday.
We could have two? Top and bottom?

kcmcleod:
> I wonder if we should only have the button on the draft type and profile pages.

I would keep it as it helps people who want to get started. Otherwise how do you create a new group/team?

kcmcleod:
now using generic enquiries@bioschemas.org for all buttons as https://github.com/BioSchemas/specifications/issues/350

AlasdairGray:
We need to consider what to do with group links back to profiles and types. In particular, since we now don't have draft versions of current types, there are several broken links. For example, Beacons no longer has a current profile, only a draft. I think every group will be broken in some way.

PR #230, opened 2019-09-12
adding text at top of profile & type to make the github link easier to find

kcmcleod:
Adding text at top of profile & type page to make the github link easier to find

### DIFF
--- a/_includes/profile_start.html
+++ b/_includes/profile_start.html
@@ -23,7 +23,7 @@
 <h3><strong>Version:</strong> {{current_version}} ({{current_pubDate | date_to_long_string }})
 <h2>{{current_subtitle}}</h2>
 <br />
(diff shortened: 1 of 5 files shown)


AlasdairGray:
How about just providing the link to the GitHub issue rather than saying where the link to it is?

kcmcleod:
> How about just providing the link to the GitHub issue rather than saying where the link to it is?

It is an attempt to make them aware of both the tabs in general and the GitHub issue links. Alan skipped right over the tabs when trying to use the pages. 
Providing a link draws attention to the GitHub, but not the tabs.

AlasdairGray:
I have always found the tabs too subtle. I've darkened the text of the tabs, but perhaps there are other visual things we can do to make them more obvious.

I found the indirection of being told where to find the link to add an issue unsatisfactory. (I had the voice of Carole saying, why do I have to look for something on the page to click. Why can't the link just be here?)

PR #234, opened 2019-10-02
BEACONS ONLY attempt at new people pages

kcmcleod:
Exploring options for people page. For BEACONS ONLY we have _groups/BeaconsPeople.md that contains group members (NOT LEADERS). This file is used to populate the group page and also the new people page:
* http://127.0.0.1:4000/groups/Beacons/
* http://127.0.0.1:4000/people/

I'm not convinced this is a better approach, see what you think. I see the following issues:
- you are still using md not something simpler like csv. The problem with csv is that it has to go in the _data folder and thus may be harder to find.
- rather than just editing 1 file (Beacons.md) the group leader now has to edit Beacons.md and BeaconsPeople.md.
- as there are no longer md files for each person there are no links to the person. All the info has to be displayed on the page.

Outstanding questions:
- what do we do with leaders? 
- what information do we want associated with each person. Currently I've kept it very simple (Name, Affiliation) but, depending on how complex you wish the md to be, we can add more info. There is no point in making it complicated as the group leaders won't add it and we can't really display it unless we have dedicated pages for each person.

Advantages
- by updating the people page from the list of people in each group these pages remain in sync and there is no need for extra work for us. It is up to the group leaders to maintain their lists.

### DIFF
--- a/_groups/Beacons.md
+++ b/_groups/Beacons.md
@@ -27,10 +27,4 @@ draft_specifications:
   [
     'Beacon'
   ]
-
-members:
-  - SusheelVarma
-  - AndraWaagmeester
-  - MichaelBaudis
-  - RafaelJimenez
 ---
--- a/_groups/BeaconsPeople.md
+++ b/_groups/BeaconsPeople.md
@@ -1,0 +1,10 @@
+---
+collection: groups
+type: members
+name: Beacons
+members:
+- Susheel Varma, Elixir Hub
+- Andra Waagmeester
+- Michael Baudis, University of Zurich & Swiss Institute of Bioinformatics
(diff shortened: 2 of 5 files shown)


AlasdairGray:
How would things look if we have someone active in two groups? Would they be displayed twice on the people page?

kcmcleod:
> How would things look if we have someone active in two groups? Would they be displayed twice on the people page?

Listing is per group. Thus if you are active in multiple groups you are listed multiple times.

ljgarcia:

AlasdairGray:
This was to explore some ideas which are now very stale. The outcome is that if we want to do something better with our people pages then we need to move to a CSV file in the `_data` directory.

PR #235, opened 2019-10-15
Added Guide to Pharmacology MolecularEntity

AlasdairGray:
Guide to Pharmacology have added MolecularEntity markup to their ligand pages (there are errors in the markup but this will be fixed in due course).

### DIFF
--- a/_liveDeploys/liveDeploy.md
+++ b/_liveDeploys/liveDeploy.md
@@ -78,7 +78,7 @@ list:
     name: IUPHAR/BPS Guide to PHARMACOLOGY
     highlight: over 9,800 ligands
     example_URL: https://www.guidetopharmacology.org/GRAC/LigandDisplayForward?ligandId=8299
-    resource_URL: https://www.guidetopharmacology.org/GRAC/LigandDisplayForward?ligandId=8299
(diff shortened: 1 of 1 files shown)


_liveDeploys/liveDeploy.md
```diff
@@ -74,6 +74,16 @@ list:
     bsc_ver: 0.1
     comments:
     node: UK
+-
+    name: IUPHAR/BPS Guide to PHARMACOLOGY
+    highlight: over 9,800 ligands
+    example_URL: https://www.guidetopharmacology.org/GRAC/LigandDisplayForward?ligandId=8299
+    resource_URL: https://www.guidetopharmacology.org/GRAC/LigandDisplayForward?ligandId=8299
```
kcmcleod: you are changing the meaning/use of resource_URL. It normally points to the home page of the resource. Is this deliberate?

AlasdairGray: Probably best if we are consistent with other descriptions. I guess we may need to think about the page layout at some point since it would be nice to be able to click to the page with the markup on as well as the resource home page.

AlasdairGray: Now fixed

PR #250, opened 2019-12-02
Live deploy improvements

AlasdairGray:
Adding feature requested in [issue 358](https://github.com/BioSchemas/specifications/issues/358) and made various improvements to the table.

Live deploy list now includes link to example page as deployed, structured data testing tool, and rich snippet generation tool. In the future we can add a Bioschemas valdator as well.

This PR also fixes the problem with the table width jumping around.

### DIFF
--- a/liveDeploys/elixir.html
+++ b/liveDeploys/elixir.html
@@ -18,7 +18,7 @@ description: |-
     <div class="container-flow">
         <div class="row">
 
-            <div class="col-md-8">
+            <div class="col-md-10">
     <section class="live-deploy-table">
     	<h5>Elixir resources using Bioschema's markup</h5>
         <ul style='text-align: left;'>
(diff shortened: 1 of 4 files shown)


liveDeploys/elixir.html
```diff
@@ -29,28 +26,30 @@ <h5>Elixir resources using Bioschema's markup</h5>
             <thead>
                 <tr>
                     <th>Name</th>
-                    <th>Profile</th>
-                    <th>Profile Version</th>
+                    <th class="profile-column">Profile</th>
+                    <th class="structured-data-column">Profile Version</th>
+                    <th class="structured-data-column">Page</th>
```
ljgarcia: Is page a web page? Not sure the column title is the right one

liveDeploys/index.html
```diff
@@ -27,16 +26,18 @@ <h5>{{liveDeploys.description}}</h5>
             <thead>
                 <tr>
                     <th>Name</th>
-                    <th>Profile Version</th>
+                    <th class="structured-data-column">Profile Version</th>
```
ljgarcia: Do not we need here the profile column as in the previous modified file (elixir.html)?

liveDeploys/elixir.html
AlasdairGray: Yes it is the web page with the deployed markup. Happy to rename to web page if you think that is clearer

liveDeploys/index.html
AlasdairGray: No. The live deploys page is grouped by profile in the concertina table. On the elixir page they are grouped by node which is why we need the profile column as well.

liveDeploys/elixir.html
ljgarcia: It is probably clear from the content so let's keep it short as it is now

ljgarcia:
@AlasdairGray It looks ready to go

PR #253, opened 2020-02-14
Create page for new profile TaxonName

frmichel:
Hi, just uploaded the yaml file generated by GoWeb from [Google Sheet](https://docs.google.com/spreadsheets/d/1cQPR7KGU0fOHXiRoaJpahEgrX9i6SIc36X8cupfg6_4/edit#gid=1483018794).

I have also committed an example here: https://github.com/BioSchemas/specifications/tree/master/TaxonName

Franck.

### DIFF
--- a/_profiles/TaxonName/0.1-DRAFT-2020_02_14.html
+++ b/_profiles/TaxonName/0.1-DRAFT-2020_02_14.html
@@ -1,0 +1,201 @@
+---
+name: TaxonName
+
+previous_version: 0.1-DRAFT-2020_02_14
+previous_release:
+
+status: revision # Change to revision if you are working on an update
+spec_type: Profile
+group: biodiversity
+use_cases_url: /useCases/TaxonName/
+cross_walk_url: https://docs.google.com/spreadsheets/d/1f8ovXRF_T_y8CR1ThHgKkF9NjnpLkIoJ_UOlPQwD6DA
+gh_tasks: https://github.com/BioSchemas/specifications/labels/type%3A%20Taxon
+live_deploy: /liveDeploys/
+
+parent_type: TaxonName
+hierarchy:
+- Thing
+- Taxon
+
+# spec_info content generated using GOWeb
+# DO NOT MANUALLY EDIT THE CONTENTspec_info:
+  title: TaxonName
+  subtitle: 'Bioschemas profile for describing a name of a biological taxon, may it
+    be valid (zoology) / accepted (botany) or not '
+  description: This profile povides a way to denote the scientific names of a taxon.
+    The taxon accepted/valid name is linked to the taxon with property scientificName
+    while the synonyms are linked with property alternateScientificName.
+  version: 0.1-DRAFT
+  version_date: 20200214T142112
+  official_type: http://schema.org/TaxonName
+  full_example: https://github.com/BioSchemas/specifications/tree/master/TaxonName/examples/
+mapping:
+- property: additionalType
+  expected_types:
+  - URL
+  description: An additional type for the item, typically used for adding more specific
+    types from external vocabularies in microdata syntax. This is a relationship between
+    something and a class that the thing is in. In RDFa syntax, it is better to use
+    the native RDFa syntax - the 'typeof' attribute - for multiple types. Schema.org
+    tools may have only weaker understanding of extra types, in particular those defined
+    externally.
+  type: ""
+  type_url: ""
+  bsc_description: 'An equivalent TaxonName type from a well known vocabulary, e.g.
+    TDWG: http://rs.tdwg.org/ontology/voc/TaxonName#TaxonName'
+  marginality: Optional
+  cardinality: MANY
+  controlled_vocab: ""
+  example: |
+    "@context": [
+          "http://schema.org",
+          { "tn": " http://rs.tdwg.org/ontology/voc/TaxonName#" }
+      ],
+    "@type" : "TaxonName",
+    "additionalType": "tn:TaxonName"
+- property: author
+  expected_types:
+  - Organization
+  - Person
+  description: The author of this content or rating. Please note that author is special
+    in that HTML 5 provides a special mechanism for indicating authorship via the
+    rel tag. That is equivalent to this and may be used interchangeably.
+  type: schema.org
+  type_url: ""
+  bsc_description: Authorship and date information associated to this taxon name,
+    formatted as per the numenclatural rules
+  marginality: Recommended
+  cardinality: MANY
+  controlled_vocab: ""
+  example: |-
+    {
+        "name": "Delphinapterus leucas",
+        "author": "(Pallas, 1776)"
+    }
+- property: description
+  expected_types:
+  - Text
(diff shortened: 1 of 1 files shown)


AlasdairGray:
Thanks @frmichel 

I've had a look at the PR and fixed a few issues that meant that the new profile was not displaying.

Can you clarify that TaxonName is a profile over the Taxon type, or is it meant to be over a new TaxonName type?

frmichel:
> 
> 
> Can you clarify that TaxonName is a profile over the Taxon type, or is it meant to be over a new TaxonName type?

Hi @AlasdairGray. As I see it, TaxonName is to be a type a type AND a profile.

AlasdairGray:
Thanks @frmichel. At the moment, the draft profile claims to be over the type Taxon so we'll need to change that.

Did you see my [question in issue 309](https://github.com/BioSchemas/specifications/issues/309#issuecomment-590960957) about TaxonName inheriting from CreativeWork? This is based on the [comment](https://github.com/BioSchemas/specifications/issues/309#issuecomment-491269932) by @qgroom in issue 309.

> A scientific Latin name for an organism is an entity in of itself. It has an author, a protologue, a date of publication and one or more type specimens. It is not however a taxon. A taxon can have four legs, but doesn't have an author.

I will ask the Leadership Group about the creation of a new profile and type.

frmichel:
Hi Alasdair,

Le 26/02/2020 à 14:40, Alasdair Gray a écrit :
>
> Thanks @frmichel <https://github.com/frmichel>. At the moment, the 
> draft profile claims to be over the type Taxon so we'll need to change 
> that.
>
Certainly my mistake, sorry about that.
>
> Did you see my question in issue 309 
> <https://github.com/BioSchemas/specifications/issues/309#issuecomment-590960957> 
> about TaxonName inheriting from CreativeWork? This is based on the 
> comment 
> <https://github.com/BioSchemas/specifications/issues/309#issuecomment-491269932> 
> by @qgroom <https://github.com/qgroom> in issue 309.
>
>     A scientific Latin name for an organism is an entity in of itself.
>     It has an author, a protologue, a date of publication and one or
>     more type specimens. It is not however a taxon. A taxon can have
>     four legs, but doesn't have an author.
>
Indeed that seems reasonable, from y point of view at least. People from 
the domain would probably be in a better posiition to confirm this.

AlasdairGray:
Sorry for the delay in catching up with this. I'm afraid that I'm going to put this PR into draft status as it needs a few changes based on the [0.5-DRAFT example](https://github.com/BioSchemas/specifications/blob/master/Taxon/0.5-DRAFT_examples/Delphinapterus%20leucas_jsonld_0.5_full.json).
- [ ] Management approval for creation of TaxonName type and profile
- [ ] Creation of TaxonName type as a subtype of CreativeWork
  - [ ] taxonRank is a property to be added to TaxonRank
  - [ ] scientificName property on Taxon has TaxonRank as a Range option

AlasdairGray:
Steering council met yesterday and approved the creation of a TaxonName profile. With regard to the type, they suggested trying to use the `about` property to avoid the need to create a new type.

frmichel:
> With regard to the type, they suggested trying to use the `about` property to avoid the need to create a new type.

What is the suggestion exactly wrt. using the about` property?

AlasdairGray:
The suggestion was that instead of introducing `taxonRank`, that the same effect could be achieved with the `about` property, something like
```json
{
  "@context": [
        "http://schema.org",
        { "tn": "http://rs.tdwg.org/ontology/voc/TaxonName#" }
    ],
  "@type" : "TaxonName",
    "additionalType": "tn:TaxonName",
    "name": "Delphinapterus leucas",
    "author": "(Pallas, 1776)",
    "about": [
      "http://rs.tdwg.org/ontology/voc/TaxonRank#Species",
      "http://www.wikidata.org/entity/Q7432",
      "species"
    ]
}
```

Is this too weak an assertion?

frmichel:
That seems a bit awkward to me. A taxon is not _about_ a rank but it has/is associated with a rank.

Plus, the taxonRank property was already defined for the Taxon type, can we not just reuse the same one for TaxonName?

ljgarcia:
If having a TaxonName as a new type extending CreativeWork rather than as a profile makes more sense for the biodiversity community, then it should be a new type.

AlasdairGray:
I've updated the Taxon and TaxonName types and profiles as discussed in this thread. @frmichel it would be good if you could verify that I've got all the changes followed through correctly.

There is just the issue of how the TaxonName should link back to the Taxon. I put a discussion of this on the [biodiversity channel](https://bioschemas.slack.com/archives/C01535HMCUS/p1591708782002600) in slack.

frmichel:
Hi @AlasdairGray, just a few remarsk:

In type Taxon (I'm looking at https://github.com/BioSchemas/bioschemas.github.io/blob/taxonname_profile_draft_0.1/_types/Taxon/0.4-DRAFT.html):
- I think we agreed that alternativeScientificName should be spelled alternateScientificName to mimic what already exists in schema.org (see [discussion](https://github.com/BioSchemas/specifications/issues/309#issuecomment-512840755) although this is not the british English way
- the range of scientificName  and alternateScientificName should be URL or TaxonName, but not text. If we want to specify a text for the name, then we should use name/alternateName instead.

In type TaxonName: 
- we could use schema:about as an optional property to link the name back to the Taxon it is used by.
- I'm not sure where Bioschemas' specific properties description should go: for author and isBasedOn we have amended the description, this should go in the profile, not the type, right?

AlasdairGray:
> In type Taxon (I'm looking at https://github.com/BioSchemas/bioschemas.github.io/blob/taxonname_profile_draft_0.1/_types/Taxon/0.4-DRAFT.html):
> 
> * I think we agreed that alternativeScientificName should be spelled alternateScientificName to mimic what already exists in schema.org (see [discussion](https://github.com/BioSchemas/specifications/issues/309#issuecomment-512840755) although this is not the british English way

Sorry that is my mistake. We should follow the schema.org conventions. Changes made.

> * the range of scientificName  and alternateScientificName should be URL or TaxonName, but not text. If we want to specify a text for the name, then we should use name/alternateName instead.

Text is included in the type to follow Schema.org conventions. We can refine that in the profile.

> 
> In type TaxonName:
> 
> * we could use schema:about as an optional property to link the name back to the Taxon it is used by.
> * I'm not sure where Bioschemas' specific properties description should go: for author and isBasedOn we have amended the description, this should go in the profile, not the type, right?

Both of these would be changes in the profile rather than the type. Are these changes that you want to make now before we put the 0.1-DRAFT on the website?

frmichel:
> Both of these would be changes in the profile rather than the type. Are these changes that you want to make now before we put the 0.1-DRAFT on the website?

The comment on isBasedOn and author are ok.

About the schema:about, to me this is a more recent request.  I'm not sure if we finally agreed on adding it, was it discussed with the committee?

AlasdairGray:
The use of `about` was a spur of the moment thought. I would suggest that we publish as is and see if there is a need for the link back from TaxonName to Taxon.

PR #254, opened 2020-02-24
EGA have implemented latest release

AlasdairGray:
EGA have updated their markup to conform with version 0.3 of Dataset and DataCatalog. They also have over 5,300 datasets with markup on them.


petrosss:
everything works fine.

PR #264, opened 2020-04-02
Fix Issue 381

AlasdairGray:
This PR fixes https://github.com/BioSchemas/specifications/issues/381 whereby the draft profiles on the [profiles page](https://bioschemas.org/profiles) link to the latest release rather than the latest draft.

I've also ensured that the logic is applied to the types page as well.


AlasdairGray:
> I am not sure what is the link you refer to. Could you please provide an example?

If you click on the Draft tab and the click on the Protein profile, it takes you to the latest published version rather than the latest draft version. This fix means that from the Draft tab you are taken to the latest draft version.

PR #265, opened 2020-04-02
Add use cases for scholarly publications profiles

ljgarcia:
Solving the issue raised at https://github.com/BioSchemas/specifications/issues/379

### DIFF
--- a/_profiles/Journal/0.1-DRAFT-2019_01_29.html
+++ b/_profiles/Journal/0.1-DRAFT-2019_01_29.html
@@ -7,7 +7,7 @@ previous_release:
 status: revision
 spec_type: Profile
 group: biotea2bioschemas
-use_cases_url: https://docs.google.com/document/d/1ZhoYDIx5DP4q1L8k6DAL3EvFuEbTnD7P981Zav_ipuE
+use_cases_url: /useCases/Journal/
 cross_walk_url: [
   "https://docs.google.com/spreadsheets/d/1ecj0bzyRO3tSTyHrKtDGfBwC5bIhgRCmv3yR8ylR_G4", 
   "https://drive.google.com/drive/folders/1CgEyta4d7vFfYT7r7w9HtLZ4kAidDmJI"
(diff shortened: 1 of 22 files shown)


ljgarcia:
Hi @AlasdairGray any news on this PR? Thanks

ljgarcia:
@AlasdairGray I went through the use cases for the release profiles and could not find an example similar to what you mentioned "Alice is searching for publications about protein ACE2?, and then state how Bioschemas will help" so I went a bit blind here but still tried to address your comment.

AlasdairGray:
Thanks for doing that. I'll try to review it later tonight. I'm pushing you in this direction as it is what we are going to require when the new governance framework is in place and it would be good to have an example to point to.

AlasdairGray:
I've tidied up some of the phrasing. I hope you find it acceptable.

ljgarcia:
@AlasdairGray this PR corresponding to [issue 379](https://github.com/BioSchemas/specifications/issues/379) was already merged but the content is not available on the website. See for instance [scholarly article  use cases](https://bioschemas.org/useCases/ScholarlyArticle/). Any idea what could have happen? Will we need to recover the branch, update it and merge it again?

AlasdairGray:
Looks like there is a filename issue. The link on the website goes to https://bioschemas.org/useCases/ScholarlyArticle/ whereas the use case seems to be in the file https://bioschemas.org/useCases/ScholarlyPublications/

ljgarcia:
@AlasdairGray It is a redirection issue. Funny thing, when I run it locally, it works well.

These are the redirections on https://bioschemas.org/useCases/ScholarlyPublications/ 
- "useCases/ScholarlyPublications/"
- "useCases/Journal/"
- "useCases/PublicationIssue"
- "useCases/PublicationVolume"
- "useCases/ScholarlyArticle"
- "useCases/SemanticAnnotation"

The first one is redundant and should be removed, right? Other than that I do not see anything else that could be breaking the redirection on the live version. Should I create a PR removing this redirection and see whether it fixes the problem? As I mentioned, it works well locally.

ljgarcia:
@AlasdairGray Could you please check again? The links are working for me from the profile list page and the profile pages as well. Thanks

AlasdairGray:
Everything seems to be working fine.

PR #266, opened 2020-04-03
Add UniProt Dataset to live deploys

AlasdairGray:


AlasdairGray:
The way the site is currently configured we can only have one node associated with a resource

PR #267, opened 2020-04-03
Fix issue 408

ljgarcia:


ljgarcia:
Only the logged change should be there. I updated accordingly.

PR #282, opened 2020-05-20
Publish Phenotype 0.2 draft

AlasdairGray:
Makes the 0.2-DRAFT of the Phenotype profile developed at the ELIXIR InterOp virtual All Hands Meeting available to the website


_profiles/Phenotype/0.2-DRAFT.html
```diff
@@ -0,0 +1,343 @@
...
+  marginality: Recommended
+  cardinality: ONE
+  controlled_vocab: ""
+  example: |
+    //Using a URL directly (looks like a text)
+    {
+      "@type": "Phenotype",
+      "image": "http://www.ebi.ac.uk/mi/impc/phenotype-archive/media/images/900/M00571072_00021593_download_tn_large.jpg"
+    }
+- property: isAssociatedWith
```
ljgarcia, line 185: This property does not exist on the corresponding type, see https://bioschemas.org/types/Phenotype/

_profiles/Phenotype/0.2-DRAFT.html
```diff
@@ -0,0 +1,343 @@
...
+    {
+      "@type": "Phenotype",
+      "anatomicalLocation": {
+                    "@type": "DefinedTerm",
+                    "name": "limb digit",
+                    "inDefinedTermSet": "http://purl.obolibrary.org/obo/emapa.owl",
+                    "termCode": "EMAPA_32725"
+            }
+    }
+- property: associatedDisease
```
ljgarcia, line 98: This property does not exist on the corresponding type, see https://bioschemas.org/types/Phenotype/

AlasdairGray:
Good catch, I had forgotten about those changes being needed to the underlying type.

_profiles/Phenotype/0.2-DRAFT.html
AlasdairGray: Resolved on BioChemEntity branch of schemaorg repo. Need to copy across to website.

_profiles/Phenotype/0.2-DRAFT.html
AlasdairGray: Is this property too generic? It is not a property that we already have defined. What does an association mean? Is this equivalent to a `rdfs:seeAlso`?

AlasdairGray: @GiuliaBabbi what were your thoughts with the `isAssociatedWith` property?

_profiles/Phenotype/0.2-DRAFT.html
ljgarcia: Do you mean a branch on https://github.com/schemaorg/schemaorg? I did not find it, could you please provide a URL? thanks

_profiles/Phenotype/0.2-DRAFT.html
ljgarcia: I think bioChemInteraction in BioChemEntity follows the same idea that isAssociatedWith in Phenotype. Both of them are broad relations that should be used if nothing more appropriate is provided.

GiuliaBabbi: Actually isAssociatedWith was already a property in google sheet draft 1.0 on which I worked during the F2F, I just switched the marginality to recommended. In my mind, it was the preferential way to associate a BioChemEntity (genes, proteins..)

ljgarcia: The previous version of Phenotype was done during the BioHackathon Europe 2018, maybe @ficolo remembers something about isAssociatedWith?

ljgarcia: gene-phenotype association is the way it will be described by people working with phenotypes. We are changing it to bioChemAssociation in case it helps, similar to the idea behind bioChemInteraction for the BioChemEntity type. And we improved the description

ljgarcia:
@AlasdairGray some additional changes after a discussion with @GiuliaBabbi , please have a look. And, if you let me know what the process to update the type is, I am happy to do so. Thanks.

_profiles/Phenotype/0.2-DRAFT.html
AlasdairGray: Yes, that is the place. I've made the necessary changes

AlasdairGray:
@ljgarcia I've updated the type to include the properties specified in the profile

PR #312, opened 2020-08-12
Added the BridgeDb mapping databases.

egonw:


egonw:
Yes, I know. that will follow. Still manual work, but I'll actually will write some scripts to automate getting this metadata from Zenodo and/or Figshare based on the DOI :) 

Oh, a fair-er world is coming :)

PR #319, opened 2020-08-31
Added missing redirect

AlasdairGray:
Ensures that the redirect patter with the trailing `/` also works


ljgarcia (approved):
Works fine for me

PR #324, opened 2020-10-01
Add call for Steering Committee

AlasdairGray:
Replaces the banner about new types/profiles with a call for the steering council.


ljgarcia (commented):
The form has more roles than those listed in the governance document. Is that ok?

AlasdairGray:
We'll discuss during the management call, but I suspect that we can get this on the site and do a minor modificaiton to the form.

PR #329, opened 2020-10-12
Convert all uses of BioSchemas to Bioschemas

AlasdairGray:
Follow on from the interaction with @martin-nc, I've done a search and replace to eliminate any usage of `BioSchemas` in the website; replacing them all with `Bioschemas`.

I've tested this on my local deployment and checked that the links to the repository work (at least some of them). 

@ljgarcia please can you also verify that this works for you and test some of the links.

### DIFF
--- a/GSoC/index.html
+++ b/GSoC/index.html
@@ -99,7 +99,7 @@ This web page gives an overview of project ideas we are offering to students who
 <h4>Difficulty</h4>
 <ul><li>Hard</li></ul>
 <h4>References</h4>
-<ul><li><a href="https://github.com/BioSchemas/bioschemas-nutch-indexer">https://github.com/BioSchemas/bioschemas-nutch-indexer</a></li></ul></ul>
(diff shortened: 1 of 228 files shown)


alaninmcr:
When I open a tab to bioschemas.org - the name that appears for the tab is BioSchemas - with an uppercase S.
bioschemas-tab

alaninmcr:
I think the tab name comes from

<title>
  			BioSchemas
  		
  </title>

AlasdairGray:
@alaninmcr is that on the live site or on a local deployment of this branch? If the former then that is what this PR corrects. If the latter then I've missed something.

When I run this branch locally I see the attached.

Screen Shot 2020-10-12 at 16 04 44
Screen Shot 2020-10-12 at 16 04 36

alaninmcr:
@AlasdairGray It was on the deployed site

AlasdairGray:
Thanks for confirming @alaninmcr

PR #332, opened 2020-10-15
Add details of where Bioschemas has been recommended for use

AlasdairGray:
Adding onto the home page details of where Bioschemas has been recommended for use to help improve the findability of resources.

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,7 +31,7 @@ call-date: 2020-10-26T15:00GMT
                 <p>Bioschemas is making two main contributions:</p>
                 <ol>
                   <li>Proposing <a href="/types">new types and properties</a> to Schema.org to allow for the description of life science resources.</li>
-                  <li><a href="/profiles">Profiles</a> over the Schema.org types that identify the essential properties to use in describing a resource.</li>
+                  <li>Defining usage <a href="/profiles">profiles</a> over the Schema.org types that identify the essential properties to use in describing a resource.</li>
                 </ol>
                 <p>
                 Bioschemas started as a community effort in November 2015. It operates as an open community initiative with <a href="/people/">representatives</a> from a wide variety of institutions. You are welcome to <a href="/howtojoin/">join the community</a>.</p>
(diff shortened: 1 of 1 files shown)


ljgarcia:
* I suggest to make it explicit it is not only about data.
Text
> Bioschemas aims to improve the Findability of data  in the life sciences. 
Not only data, we might want to go for digital research objects, digital objects, research objects or something like that as we cover data, software, workflow, training materials.

* I suggest being more specific about findability benefits
Text
> metadata means that you meet some of the Findability criteria of the FAIR Data Principles
Rather than "some" or as a complement, we can include F2 for sure while encouraging F1, F3 and F4. We could also mention some additional effects regarding I1 (particularly on the metadata) and R1.3 while also encouraging R1.1 and R1.2

AlasdairGray:
> * I suggest to make it explicit it is not only about data.
>   Text
> 
> > Bioschemas aims to improve the Findability of data  in the life sciences.
> > Not only data, we might want to go for digital research objects, digital objects, research objects or something like that as we cover data, software, workflow, training materials.

I've tried to address this in 07b675b023349ba78cbb8445791df1b073cf5b15

AlasdairGray:
> * I suggest being more specific about findability benefits
>   Text
> 
> > metadata means that you meet some of the Findability criteria of the FAIR Data Principles
> > Rather than "some" or as a complement, we can include F2 for sure while encouraging F1, F3 and F4. We could also mention some additional effects regarding I1 (particularly on the metadata) and R1.3 while also encouraging R1.1 and R1.2

I think this is getting too detailed for the front page of the website. Perhaps better placed in the training materials. We could include a link to where a reader can find these details.

ljgarcia:
@AlasdairGray I like the changes, thanks!

ljgarcia:
> 
> 
> > * I suggest being more specific about findability benefits
> >   Text
> > 
> > > metadata means that you meet some of the Findability criteria of the FAIR Data Principles
> > > Rather than "some" or as a complement, we can include F2 for sure while encouraging F1, F3 and F4. We could also mention some additional effects regarding I1 (particularly on the metadata) and R1.3 while also encouraging R1.1 and R1.2
> 
> I think this is getting too detailed for the front page of the website. Perhaps better placed in the training materials. We could include a link to where a reader can find these details.

Indeed

ljgarcia (approved):
Looks good to me

PR #333, opened 2020-10-19
Champions page

AlasdairGray:
Added a page describing the Bioschemas Champion role and listing those who have agreed to take on this role.


ljgarcia (approved):
Looks fine

PR #334, opened 2020-10-19
Remove test files

AlasdairGray:
Two test files were inadvertently committed and merged to master as part of the BioSchemas elimination patch. The files should not have been included but have no effect on the visible website.

This patch removes the two test files to avoid confusion and potential file clashes in the future.


petrospaps (approved):
Hi Alasdair, looks good to me.

PR #337, opened 2020-10-28
Add COVID-19 data portal to live deploy list

AlasdairGray:
The COVID-19 data portal now has DataCatalog and Dataset markup on their pages and are therefore eligible for listing on the live deploy page.


ljgarcia (approved):
This is good news

PR #339, opened 2020-11-02
Update Training Group Leads

AlasdairGray:
Update the training group leads. Updated the display to include former training group leads and associated dates.


ljgarcia (approved):
Looks fine

PR #340, opened 2020-11-03
Adding details of known sitemaps

AlasdairGray:


danbri:
Thanks!

petrospaps:
Hi Alasdair, the sitemaps are added fine, but they are not visible to the user in the live deploys page. Is this the intention or am I looking to the wrong page?

AlasdairGray:
At the moment there is no code to make the sitemaps display. This is an enhancement that we will need to make in the future. However, having the sitemaps in the live deploy data file is still useful.

petrospaps (approved):
Then everything looks good.

PR #343, opened 2020-11-19
Fix Issue 457: various improvements to the display of a type

AlasdairGray:
See https://github.com/BioSchemas/specifications/issues/457 for details.

Fixes:
- Legend text improved (extracted into common file for both types and profiles)
- CD legend text only shown on profile pages
- Profile links show on type page

@ljgarcia can you check the text of the legend text

### DIFF
--- a/_includes/spec-table-key.html
+++ b/_includes/spec-table-key.html
@@ -1,0 +1,11 @@
+
+<h5>Key to specification table</h5>
+<ul>
+  <li><em style="color: #0B794B;"><strong>Green</strong></em> properties/types are proposed by Bioschemas, or indicate proposed changes by Bioschemas to Schema.org</li>
+  <li><em style="color: #990000;"><strong>Red</strong></em> properties/types exist in the core of Schema.org</li>
+  <li><em style="color: #0000CC;"><strong>Blue</strong></em> properties/types exist in the pending area of Schema.org</li>
+  <li><em style="color: #454547;"><strong>Black</strong></em> properties/types are reused from external vocabylaries/ontologies</li>
+</ul>
+{% if page.spec_type == 'Profile' %}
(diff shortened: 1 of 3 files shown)


_includes/spec-table-key.html
```diff
@@ -0,0 +1,11 @@
+
+<h5>Key to specification table</h5>
+<ul>
+  <li><em style="color: #0B794B;"><strong>Green</strong></em> properties/types are proposed by Bioschemas, or indicate proposed changes by Bioschemas to Schema.org</li>
+  <li><em style="color: #990000;"><strong>Red</strong></em> properties/types exist in the core of Schema.org</li>
+  <li><em style="color: #0000CC;"><strong>Blue</strong></em> properties/types exist in the pending area of Schema.org</li>
+  <li><em style="color: #454547;"><strong>Black</strong></em> properties/types are reused from external vocabylaries/ontologies</li>
```
ljgarcia, line 7: Typo: vocabylaries --> vocabularies

ljgarcia: Are the black ones allowed in Bioschemas profiles? I thought we wanted all the properties to be in schema.org or Bioschemas but not outside

ljgarcia (changes requested):
Mainly for the typo. The question about the black category legend could be left for later if this PR needs quick approval

_includes/spec-table-key.html
AlasdairGray: For properties that are going to be used by the vast majority of use cases then we should push for properties to be included into Schema.org. However, there are instances of edge cases where just using the domain term makes sense. One example is the plant community want to state the type of growth facility.

ljgarcia (approved):

PR #356, opened 2021-01-19
Enable link to live deploys

AlasdairGray:
Changed setting so that the link for live deploys takes you to the live deploy page.


ljgarcia (approved):
Works fine for me

PR #358, opened 2021-01-25
Tutorials

ljgarcia:
Brings information from https://bioschemas.gitbook.io/training-portal/ so we can add structured data.
So far is only a copy&paste from the GitBook space.


AlasdairGray (approved):
I fixed a few typos that I spotted. 

Happy for this to be made live in its current form with the idea that it gets revised and improved as we go.

PR #359, opened 2021-01-26
Steering council update

AlasdairGray:
Add Alban and Ivan to the Steering Council.

Changed layout of Steering Council information to being a table.


ljgarcia (approved):
Looks good

PR #362, opened 2021-01-29
Show the latest release number on live deploys headers

AlasdairGray:
Adds the latest release number (or draft number if there hasn't been a release yet) to the header rows in the live deploy table. (As discussed during the Steering Committee meeting yesterday.)


ljgarcia (commented):
I saw the addition of the version under the profile name, that looks good. 
However, I was expecting to see "dc:conformsTo" in the json-ld markup to figure out the Bioschemas profile version used but did not find it on several of the pages listed on the live deploys. How can be sure what is the version used?

AlasdairGray:
@ljgarcia we do not control the markup on the pages, but we have been recommending for more than a year in our tutorials that `dct:conformsTo` should be included in the markup and the BioschemasGenerator includes this as part of the template markup together with `@context`, `@id`, and `@type`. These recommendations should be made more prominent in our online tutorial materials. However, I don't think that it should stop this PR being accepted.

ljgarcia (approved):

PR #363, opened 2021-02-02
Tutorial on checking deployed markup

AlasdairGray:
This is a first version of a tutorial on checking your deployed markup.


ljgarcia (approved):
It looks good. I suggest to add an example with https://search.google.com/test/rich-results as https://search.google.com/structured-data/testing-tool is supposed to be deprecated and replaced by the former. This does not have to be done as part of this PR so approving now with the current content.

PR #365, opened 2021-02-08
Linkfix

alaninmcr:
This makes the links in the tables correct (I think)

### DIFF
--- a/_includes/profile_start.html
+++ b/_includes/profile_start.html
@@ -83,9 +83,11 @@
      branch_name == "BioChemStructure" or
      branch_name == "BioSample" or
      branch_name == "ChemicalSubstance" or
+     branch_name == "ComputationalWorkflow" or
      branch_name == "DataRecord" or
      branch_name == "DNA" or
      branch_name == "Enzyme" or
+     branch_name == "FormalParameter" or
      branch_name == "Gene" or
      branch_name == "LabProtocol" or
      branch_name == "MolecularEntity" or
--- a/_includes/specification_table.html
+++ b/_includes/specification_table.html
@@ -66,7 +66,7 @@
             {% for prop in props_list %}
             <tr>
                 {% if prop.type == "bioschemas" %}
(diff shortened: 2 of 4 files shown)


AlasdairGray:
@alaninmcr which links in particular should this fix?

alaninmcr:
In the HTML page for the profile of ComputationalWorkflow, there are links to FormalParameter. Those links tried (and failed)) to go to schema.org because FormalParameter (and ComputationalWorkflow) were not in the enumerated list of bioschemas types. Putting them into the list makes the links go to bioschemas.

AlasdairGray:
Thanks, that is what I thought. Its just that the links for the types were still not taking me to the respective type when I tried it.

alaninmcr:
Oh boo. Where do they take you to?

--
Alan

The University of Manchester belongs to the European R&D infrastructure IBISBA 1.0
The project has received funding from the European Union’s Horizon 2020
Research and Innovation Programme under grant agreement No 730976.
To find out more visit https://www.ibisba.eu

________________________________________

AlasdairGray:
On the draft ComputationalWorkflow (http://localhost:4000/types/ComputationalWorkflow/0.1-DRAFT-2020_07_21/) it takes me to `http://localhost:4000/types/ComputationalWorkflow/0.1-DRAFT-2020_07_21/#FormalParameter` instead of `http://localhost:4000/FormalParameter`.

Just tested another type (BioChemStructure) and the links there behave as I would expect.

alaninmcr:
Does it work in the profiles? Those are the ones I looked at fixing. I will look at the types now.

AlasdairGray:
Sorry I was looking at the types when you'd clearly said profile earlier in the thread, but best that we get them all sorted.

Type links in the profile and type are behaving as expected, as are property links in the type. I've fixed the behaviour of the property links in the profile.

AlasdairGray (approved):
Thanks for the fix @alaninmcr

PR #372, opened 2021-02-25
Add GBIF to Taxon live deploys

thomasstjerne:


AlasdairGray (approved):
thanks for the contribution.

You may want to change to `https://schema.org/` for your context to enable tools to process your JSON-LD.

PR #380, opened 2021-03-22
Move tutorials to the Bioschemas website

AlasdairGray:
This PR brings the content of the gitbook onto the Bioschemas website which permits us to embed markup within the content (eating our own dog food). The tutorial template ensures that a consistent approach is provided for embedding the markup and exploiting it in the display of the tutorials, and conforms to the 0.9-DRAFT of the TrainingMaterial profile.

The Getting Started menu item has been updated to link to key tutorials in the collection.


ljgarcia:
@albangaignard I work on some of the files changes as part of this PR so maybe better if someone else is the reviewer. Could you please have a look? Thanks

AlasdairGray:
> @albangaignard I work on some of the files changes as part of this PR so maybe better if someone else is the reviewer. Could you please have a look? Thanks

Good point @ljgarcia, I'd missed that when I did a quick glance today.

albangaignard:
I was first confused, since I could not find any reference to "tutorials"  in the menu "Getting started". Why not reusing the tutorial title directly as a menu item ?

albangaignard:
There a pagination issue when going back from the "Schema.org, what and why?" tutorial : 
Capture d’écran 2021-04-01 à 10 10 08

But it's working fine for the other tutorial.

albangaignard:
I could find all JSON-LD markup for the tutorials (no markup for the overview page, but that's fine since it's not a tutorial properly speaking)

AlasdairGray:
> I was first confused, since I could not find any reference to "tutorials" in the menu "Getting started". Why not reusing the tutorial title directly as a menu item ?

@albangaignard do you mean that the top level menu item should be changed from "Getting started" to tutorial, or are you suggesting that the first item "Overview" should be changed to "tutorials"?

AlasdairGray:
> There a pagination issue when going back from the "Schema.org, what and why?" tutorial :
> Capture d’écran 2021-04-01 à 10 10 08
> 
> But it's working fine for the other tutorial.

This is the first tutorial. Were you expecting the previous link to take you back to the overview of the tutorials?

albangaignard:
Ok, so maybe we could have numbered tutorials in the menu, "Tutorial 0: Overview", "Tutorial 1: Schema.org", and this would fix the navigation-pagination issue ?

AlasdairGray:
You were expecting a pointer back to the overview, that can be fixed easily. Would you also expect that the last tutorial would point back to the overview with it's next pointer?

I'd be wary about numbering the tutorials as we are likely to be adding more as we go along. Longer term we should probably move the ordering of the tutorials into a data file rather than the hard coding that we are doing at the moment.

albangaignard:
Ok, good point for not numbering the tutorials. 
So, just aving a pointer back to the overview should be enough for the moment.

AlasdairGray:
Both ends, or just the start?

albangaignard:
It seems to me that there is a progression in tutorials, so having nothing after the last tutorial is ok. Let's say just the beginning.

AlasdairGray:
Changes made. I've also updated things so that on the last page there is no next link and rephrased the link on the overview page (which is hard coded)

albangaignard (approved):
Working fine on my local copy

PR #387, opened 2021-04-01
Fix biogicalRole typo on some of the released types

AlasdairGray:
Fixes [Issue 484](https://github.com/BioSchemas/specifications/issues/484)

As this is a typo, the Steering Council have agreed that this should just be corrected without updating any version numbers. Note that the typo was not included in the version that went to Schema.org.


ljgarcia (approved):
Looks good to me

albangaignard (approved):
It works on my local copy

PR #389, opened 2021-04-14
Attempt to fix Issues 494

AlasdairGray:
Removing the trailing `/` seems to resolve the [issue](https://github.com/BioSchemas/specifications/issues/494)


ljgarcia (approved):
Works fine for me

PR #397, opened 2021-04-23
Update what_why_bioschemas.md

nsjuty:
- some tidying up
- modified text to include harvesting to EOSC, rather than solely focus on integration into schema.org


ljgarcia (approved):
Looks good

PR #403, opened 2021-05-08
Update Live Deploy page to use data in JSON format

AlasdairGray:
This is a big change to the live deploy page, and some of it is not perfect, but I believe that it is a big improvement over what we currently have. 

The existing YAML live deploy data is replaced with a JSON file that focuses on sites with profiles, i.e. there is now one entry per site and if the site has multiple profile deployments then these are nested within the site. It has also been extended to include information such as the sitemap location. A GitHub Action has been added to validate the JSON file against a JSON Schema file. This should ensure consistency in the data going forward.

The live deploy page has been updated to work with the new JSON version of the data. It provides three tabs with different views of the data: By site, By profile (emulates old live deploy page), and By ELIXIR Node (emulates old elixir live deploy page). It now provides a count of the number of sites that have deployments, and a count of the number of profile deployment (this is what we were previously reporting). The ELIXIR live deploy page has been removed and added as an extra tab, with counts for the number of ELIXIR site deployments as well as breakdown in the number of CDRs, DDRs, and RIRs that have deployments.

Note that I have replaced the use of the Rich Snippet Tool with the BMUSE deployment on the Heriot-Watt server.

Issues that can be fixed in a future iteration would be to:
- Display the page name on the By profile tab, at the moment it shows the URL; however this may be challenging in Jekyll with the limitations on the expressivity of the language
- Extra whitespace appears in the By Elixir node table
- On the By Elixir node page, split sites that come from multiple nodes to their constituent nodes. This is where I hit the limits of Jekyll as you cannot write to the array in Liquid so I was unable to update the array of live deploys by node to split apart bio.tools into the different nodes (This limitation is on the existing live deploy page)


AlasdairGray:
To help your reviewing, here are screenshots of the different tabs
Screen Shot 2021-05-14 at 15 11 50
Screen Shot 2021-05-14 at 15 11 58
Screen Shot 2021-05-14 at 15 12 21

albangaignard:
There is an issue with "CHD7 Database"  live deploy, the URL seems to be broken.

albangaignard:
Very nice work and great improvements !

albangaignard (changes requested):
- CHD7 Database URLs should be fixed
- same for http://scientificdata.isa-explorer.org 
- loading of https://synbiohub.org/public/igem/BBa_K165005/1 is very long 
- same for https://inpn.mnhn.fr/espece/cd_nom/60878/

AlasdairGray:
> * CHD7 Database URLs should be fixed

Correct URL is https://www.chd7.org/; change has been made.

> * same for http://scientificdata.isa-explorer.org

Removed as the site is no longer working

> * loading of https://synbiohub.org/public/igem/BBa_K165005/1 is very long
> * same for https://inpn.mnhn.fr/espece/cd_nom/60878/

This is beyond our control.

egonw:
> * CHD7 Database URLs should be fixed
> * same for http://scientificdata.isa-explorer.org
> * loading of https://synbiohub.org/public/igem/BBa_K165005/1 is very long
> * same for https://inpn.mnhn.fr/espece/cd_nom/60878/

See @AlasdairGray's changes, but I would do data updates after the merge anyway, because otherwise we keep making data updates here.

AlasdairGray:
@albangaignard that is the conflict resolved by adding the `_liveDeploys/liveDeploy.md` file back into this branch. Once this PR is accepted we will then delete the file on a separate branch and PR cycle.

albangaignard (approved):
Works for me on my local copy.

PR #405, opened 2021-05-20
ProteinAnnotation profile versions

AlasdairGray:
In reviewing the protein annotation profile, I found that the last two versions, which date from the BioHackathon in 2018, have not been added to the website. This PR takes the content of the crosswalks and makes them available.


ljgarcia (approved):
Looks good

PR #410, opened 2021-05-26
Update community call date

AlasdairGray:


ljgarcia (commented):
When testing I see "Next call: 28 June, 2021 16:00 W. Europe Daylight Time" Is W. Europe Daylight Time the same as BST? If yes, then the time is correct and I will accept and merge this PR. If not, something is failing in the time display. @AlasdairGray could you please confirm? Thanks

AlasdairGray:
Something funky going on with the timezones here. I just tested it on my machine and it displays as I expect, but the time of the last meeting (front in the picture) is incorrect. All that I changed in the PR is the date, not the timezone information. So I suspect that this is not going to display correctly.
Screen Shot 2021-05-27 at 17 35 42

AlasdairGray:
@ljgarcia I've changed the time declaration to be UTC so that it will at least be correct.

ljgarcia (approved):

PR #413, opened 2021-05-28
First draft of creating a new profile tutorial

AlasdairGray:
@ljgarcia as I was creating a new profile this afternoon, I thought I would document the process as a tutorial. This PR contains the result.

I will leave it for you to decide where this should be embedded within the tutorial structure.


ljgarcia:
It looks good but might be a good idea to ask someone who has not done a new profile yet to try the steps and see if all the information is there.

AlasdairGray:
> It looks good but might be a good idea to ask someone who has not done a new profile yet to try the steps and see if all the information is there.

Perhaps we can just accept it and then revise when issues are found.

ljgarcia:
> 
> 
> > It looks good but might be a good idea to ask someone who has not done a new profile yet to try the steps and see if all the information is there.
> 
> Perhaps we can just accept it and then revise when issues are found. It is most likely to be me who is creating these anyway.

Sounds good to me

ljgarcia (approved):

PR #414, opened 2021-06-02
Replace Google's SDTT with Schema.org Markup Validator on Live Deploy page

AlasdairGray:
This addresses [#498](https://github.com/BioSchemas/specifications/issues/498).

Google's structured data testing tool has been deprecated and validates markup against Google's internal opinion of what properties should be provided.

Schema.org have a new validator that does not impose a profile over the top of the markup. It only validates the syntax.


ljgarcia (approved):
looks fine to me

PR #418, opened 2021-06-03
Fix URL used by Schema.org validator

AlasdairGray:
While the `?` in the URL pattern worked, it ended up with a strange duplication of the URL pattern in the service used. This PR is a minor change which hopefully makes our calls conformant with the expected call pattern.


ljgarcia (approved):
working fine

PR #424, opened 2021-06-04
Add Orphanet to the list of live deploys

AlasdairGray:
Orhpanet have a deployment of the Disease profile. Needed to extend the JSON-Schema to include Disease as a valid profile name. Also updated the Disease profile to show that there is a live deployment.


ljgarcia (approved):
Looks good

PR #430, opened 2021-06-24
Fix Missing Redirects for DNA and RNA types

AlasdairGray:
[Issue 500](https://github.com/BioSchemas/specifications/issues/500) identified problems with links from Protein profile to DNA and RNA types.

Problem relates to missing redirects which are fixed in this PR.


albangaignard (approved):
works in my local copy

PR #431, opened 2021-06-25
Add BMRB to live deploy list

AlasdairGray:


albangaignard (approved):
works in my local copy

PR #433, opened 2021-06-30
Address first and second tasks on 487 issue

ljgarcia:
See https://github.com/BioSchemas/specifications/issues/487 
It adds numbers to steps in the tutorials and more descriptive titles on the menu


AlasdairGray (approved):
I've replaced intro with introduction

PR #434, opened 2021-07-01
Update next call date

AlasdairGray:
If we cancel the July meeting, then we'll update this again.


AlasdairGray:
Based on the community poll, we should skip the July meeting and have the next meeting on 23 August.

ljgarcia (approved):
Looks good to me

PR #436, opened 2021-07-01
Ensembl have Gene and DataRecord deployments

AlasdairGray:
Extends the Ensembl live deploy listing with Gene and DataRecord.

Note that the markup is visible if you inspect the rendered page in Chrome. It is not being detected by SMV or BMUSE.


ljgarcia (approved):
Looks fine to me so I am accepting this PR.
However, DataRecord is deprecated and did not make it to schema.org/pending... Do we want to include deployments of deprecated specifications? Opening an issue in this regards (https://github.com/BioSchemas/specifications/issues/524)

PR #443, opened 2021-07-24
No css reset

bedroesb:
Is this CSS reset really needed, I've never seen this. I think it is something from the past where browsers where very different from each other? I had some weird behaviors because of it. like 
```
<small>Some text</small>
```

Gets reset because of it . If I take it away everything seems fine on my browser. @AlasdairGray Could you check on yours?


AlasdairGray (changes requested):
@bedroesb can you explain where I might see differences?

Just looking through some random pages, it seems that the formatting of profile tables has changed; I've lost the white borders in the tables.
- http://127.0.0.1:4000/profiles/ChemicalSubstance/0.4-RELEASE/
- https://bioschemas.org/profiles/ChemicalSubstance/0.4-RELEASE/

bedroesb:
Ok we might want to check for these kind of changes. Reason why I would take this away is because it was overwriting some Bootstrap classes. And it is also not needed if we make correctly use of bootstrap. Visually it will normally not change (much).

bedroesb:
The example I give in the description of this PR is used in the 'news' branch for the panels on the homepage

bedroesb:
To be further tested in the 'tutorial-restyling' branch

AlasdairGray:
I agree that the change for `<small>` was an improvement, but the impact on the specification tables was a big problem.

PR #444, opened 2021-07-26
Changed to relevant example

alaninmcr:


ljgarcia (approved):
Looks good. 
I split content into two subtitles to get a TOC at the beginning.

PR #445, opened 2021-08-06
Back to top

bedroesb:
New back to top animation: 

Appears automatically using javascript when the page is scrolled beyond the size of the window + makes use of the similar styling currently present on the website

Screenshot from 2021-08-06 11-42-45


AlasdairGray (changes requested):
Same issue with styling of tables as in Pull #443 (same code removal seems to have taken place).

Really like the floating back to top. However, it doesn't seem to appear on some pages, e.g. http://127.0.0.1:4000/news/

bedroesb:
@AlasdairGray yes, same will appear with the tutorial-restyling branch, I will come with a fix for these branches and close my PR for the CSS reset removal.

bedroesb:
About the news page, that is correct, the page has not the minimum length to appear. The reasoning is that if your page is not long, there is no reason to let it appear. If you make a small window and reload the news page and scroll down, it will appear (it is calculated based on the window length)

bedroesb:
@AlasdairGray I decreased the threshold, now the user has to scroll 1/3 of its window length down for it to appear, do you which it to be sooner?

AlasdairGray:
> @AlasdairGray yes, same will appear with the tutorial-restyling branch, I will come with a fix for these branches and close my PR for the CSS reset removal.

Alternatively remove the CSS reset commit from here and this can be accepted. Then you can fix the styling in the other branch. (Unless there is something else in the styling reset that you are relying on for this back to top to appear.)

bedroesb:
that we can do too :smile:

bedroesb:
Done!

AlasdairGray (approved):

PR #450, opened 2021-08-24
Fixing mobile + link fix

bedroesb:
Possibility to fix some issues: Taking away the sticky header 

This fixes the hover/click problem + the header covering the title in mobile mode

### DIFF
--- a/_sass/_header.scss
+++ b/_sass/_header.scss
@@ -25,18 +25,4 @@
         bottom: 0.5em; 
         box-shadow: 0 4px 8px 0 rgba(0, 0, 0, 0.2);
     }
-    
-    &.stuck {
-        .hexagon {
-            padding:0.2em;
-            width: 38px;
-            position: fixed;
-            top: 0.2em;
-            left: 0.2em;
-            z-index: 300;
-            background: #fff;
-            border-radius: 50%;
-            min-width: initial;
-        }
-    }
 }
--- a/_sass/_responsive.scss
+++ b/_sass/_responsive.scss
@@ -19,13 +19,6 @@
             text-align: center;
         }
     }
-    .trigger.stuck {
-        position: fixed;
-        z-index: 100;
-        top: 0;
-        background-color: rgb(11, 121, 75);
-        width: 100%;
-    }
 }
 
 @media screen and (max-width: 700px) {
--- a/_sass/_responsive.scss
+++ b/_sass/_responsive.scss
@@ -27,8 +27,6 @@
     }
     #header_wrap {
         height: 50px;
-        position: fixed;
-        z-index: 90;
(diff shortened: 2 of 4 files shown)


AlasdairGray:
@bedroesb is the idea of removing the sticky header a temporary one, i.e. the header will be reintroduced with the handlebar upgrade?

@ljgarcia what are your thoughts on the sticky header being removed?

ljgarcia:
As there are concerns about this solution, I accepted the tutorial-restyling PR before merging this branch into it. A new PR to the master branch will be required

ljgarcia (approved):
Initial tutorial-restyling branch with agreed changes during the community meetings has been merged into master. There is not harm in merging this branch into the tutorial-restyling so I am accepting this PR although more review is needed

ljgarcia:
> 
> 
> @bedroesb is the idea of removing the sticky header a temporary one, i.e. the header will be reintroduced with the handlebar upgrade?
> 
> @ljgarcia what are your thoughts on the sticky header being removed?

I am not sure. I kind of like the sticky header as it is now. This is not a minor change, I suggest we discuss it during the September community call. @AlasdairGray @bedroesb would be ok to delay until then?

bedroesb:
hmm I would not have merged it tbh, but changed the branch where it got merged too.

ljgarcia:
@bedroesb please, feel free to reverse the merge

bedroesb:
@ljgarcia no problem, I restarted a PR from the tutorial restyling branch. You are right to discuss the fixed header thing in the next meeting!

AlasdairGray:
If the fixed header is causing problems with other aspects of the UX then I would suggest that we accept that we lose it. I've noticed that a lot of sites these days don't use fixed headers anymore.

@bedroesb is it the case that to have the fixes that you've done that we need to lose the fixed header?

bedroesb:
I will give it today one more time to fix the problems with the fixed header. I noticed that GitHub has a fixed header and does not have the problems we have + does not use the hacky solution with the headers, so there might be still hope!

bedroesb:
@AlasdairGray I found an awesome solution for the subheadings! https://stackoverflow.com/a/13067009 It works great

bedroesb:
Man I love stack overflow :D

bedroesb:
and with the new bootstrap 5 navbar we also don't have the mobile problem by default

bedroesb:
@AlasdairGray I could fix everything in the bootstrap5 branch, but it did not work with the current version ... Did some updates on that branch so it is again a little step closer towards something presentable: https://bedroesb.github.io/bioschemas.github.io

It would be nice to have a call together with @ljgarcia and you to discuss some stuff that are in html, but could be just in markdown.

PR #446, opened 2021-08-19
Moving to Bootstrap 5 + other improvements

bedroesb:
I couldn't hold myself and gave it a go:

- [x] Moving CSS and JS to an assets directory
- [x] Merging CSS and _SCSS + no stylesheet.css anymore (moved to main.scss)
- [x] Updating to bootstrap 5.1 + making use of its scss files
- [x] No popper.js, part of bootstra.bundle
- [x] Using relative links for everything
- [x] Fixing the deployment on other forks, you can preview this PR [HERE](https://bedroesb.github.io/bioschemas.github.io/) :partying_face: 
  - [x] Home links not working
- [x] New Top navigation bar purely based on bootstrap!
  - [x] Fixing the logo/title in mobile mode
- [x] No separate side mobile nav anymore -> easier to maintain
- [x] Replacing as much of the custom styling with standard bootstrap classes
- [x] Moving logos that are not part of the content to assets/img
- [x] Restructuring the default layout to make it more compatible with WCAG 2.1

Ongoing: Major refactoring -> will wait for further refactoring until the `tutorials-restyling` and `news` branch aremerged into master

TO DO:

- [x] Fixing the modals 
- [x] Fixing animation of logo
- [x] Fixing people/groups that it does not pickup index file
- [x] Publications page
- [x] Fixing Live deploys

PR #451, opened 2021-08-24
Fixing mobile + link fix

bedroesb:
Possibility to fix some issues: Taking away the sticky header

This fixes the hover/click problem + the header covering the title in mobile mode


bedroesb:
this PR is not needed anymore since it is fixed in #446

PR #458, opened 2021-09-15
Add ChEMBL Data Catalog to live deploys

AlasdairGray:


ljgarcia (commented):
It looks good but version used is missing. Any chance they can easily/quickly add it?
If not, should I go for an issue and accept this PR?

PR #459, opened 2021-09-21
Fix https://github.com/BioSchemas/specifications/issues/536

AlasdairGray:
@alaninmcr the Jekyl redirect requires both the trailing `/` and non-trailing `/` forms to be included. This should hopefully fix the problem.

As you say, local testing works but not on the server so we won't know for sure until this is deployed.


alaninmcr (approved):
Looks OK to me - but should similar changes be applied to ComputationalWorkflow?

PR #460, opened 2021-09-22
Add Bioversions to live_deployment.json

cthoyt:
The home page of the bioversions site (https://biopragmatics.github.io/bioversions/) lists several datasets. Please let me know if this is okay.

Also, this schema doesn't fully conform to this draft, so is it valid if I remove that?

### DIFF
--- a/_data/live_deployments.json
+++ b/_data/live_deployments.json
@@ -1248,14 +1248,14 @@
     },
     {
       "name": "Bioversions",
-      "keywords": ["versioning", "biomedical databases"],
       "url": "https://biopragmatics.github.io/bioversions/",
       "sitemap": "https://biopragmatics.github.io/bioversions/sitemap.xml",
       "profiles": [
         {
           "profileName": "Dataset",
-          "conformsTo": "0.4-DRAFT",
+          "conformsTo": "-",
           "exampleURL": "https://biopragmatics.github.io/bioversions/",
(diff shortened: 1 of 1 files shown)


egonw (approved):
Didn't check the JSON, but it looks good to me.

AlasdairGray:
I'm removing the keywords, as these are used for display purposes on the Bioschemas site and have a specific meaning (we really need to document that). I've also removed the conformance version and added a note in the highlight field since I don't believe that you can conform with our profile.

AlasdairGray:
I would suggest that you `Website` annotation could be changed to `DataCatalog` which is a specific type of website that aggregates datasets.

AlasdairGray:
i don't know what I've done, but my edits seem to have ended up on a separate branch https://github.com/BioSchemas/bioschemas.github.io/tree/cthoyt-patch-1

Is anyone able to fix this?

cthoyt:
You should be able to add my repo as a remote and push to it directly (do `git fetch --all` first) or edit directly through the github browser.

cthoyt:
I also updated it here

egonw:
> I've also removed the conformance version and added a note in the highlight field since I don't believe that you can conform with our profile.

Sorry, that's my bad. I put that in the template. Didn't think of it :/

AlasdairGray (approved):
Thanks for contributing your resource to our list

PR #462, opened 2021-09-22
Adding a tag icon to the tags

bedroesb:
This is a possible representation of the tags
Screenshot from 2021-09-22 16-10-50
Screenshot from 2021-09-22 16-11-25
Screenshot from 2021-09-22 16-11-17


AlasdairGray (approved):
Looks great

PR #464, opened 2021-09-27
Remove ELIXIR-PT from Live Deploys list

AlasdairGray:
The page that we have listed is no longer available. The replacement page no longer has markup embedded, so we need to remove from the live deploy list.

I have checked with E-PT and not had a response back about whether they will reenable the markup.

This fixes issue https://github.com/BioSchemas/specifications/issues/525

### DIFF
--- a/_data/live_deployments.json
+++ b/_data/live_deployments.json
@@ -281,18 +281,6 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


albangaignard (approved):
works in my local copy

PR #472, opened 2021-10-11
SequenceAnnotation 0.7-DRAFT and SequenceRange 0.1-DRAFT

AlasdairGray:
Closes issues:
- https://github.com/BioSchemas/specifications/issues/283
- https://github.com/BioSchemas/specifications/issues/468
- https://github.com/BioSchemas/specifications/issues/469

Extends release profiles:
- Gene 1.0-RELEASE -> Gene 1.1-DRAFT
- Protein 0.11-RELEASE -> Protein 0.12-DRAFT

Deprecates ProteinAnnotation, replacing it with SequenceAnnotation.

Will need to do a separate PR to catch the types up with these profiles but it is important that we get these out rather than sitting on them for longer.


AlasdairGray:
@ljgarcia as stated in the original PR text, a separate request will be needed to update the types. This is more complex since it affects types that have already been integrated into Schema.org. In fact, we need to completely reconsider how we do types. I will add this onto the Steering Council's meeting agenda.

For now, I think it is important that we allow profiles to continue to evolve. The information needed for the properties is available in the profile, i.e. property name, expected types, and description. How they fit into the hierarchy, i.e. what type they are attached to, is orthogonal to the development of the profile.

This is also not the first time that we would have profiles using types/properties that are not listed in the type section of the website.

Hopefully this convinces you enough to allow you to approve this request.

ljgarcia:
> 
> 
> @ljgarcia as stated in the original PR text, a separate request will be needed to update the types. This is more complex since it affects types that have already been integrated into Schema.org. In fact, we need to completely reconsider how we do types. I will add this onto the Steering Council's meeting agenda.
> 
> For now, I think it is important that we allow profiles to continue to evolve. The information needed for the properties is available in the profile, i.e. property name, expected types, and description. How they fit into the hierarchy, i.e. what type they are attached to, is orthogonal to the development of the profile.
> 
> This is also not the first time that we would have profiles using types/properties that are not listed in the type section of the website.
> 
> Hopefully this convinces you enough to allow you to approve this request.

@AlasdairGray yes, we can go ahead with the profiles even if the types will come later.
But, what about the typo?

>It does not affect functionality but I think the correct spelling is supersede rather than supercede. See >https://dictionary.cambridge.org/spellcheck/english/?q=Supercede

AlasdairGray:
> @AlasdairGray yes, we can go ahead with the profiles even if the types will come later. But, what about the typo?
> 
> > It does not affect functionality but I think the correct spelling is supersede rather than supercede. See >https://dictionary.cambridge.org/spellcheck/english/?q=Supercede

Sorry, I'd missed that. Spelling is now corrected. Functionality working for me.

ljgarcia (approved):
Looks good on local deploy

PR #478, opened 2021-11-02
Create Pull Release template for profile release

AlasdairGray:
The template contains the checklist of items that need to be achieved for a profile to move to release version.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/profile-release-PR-template.md
+++ b/.github/PULL_REQUEST_TEMPLATE/profile-release-PR-template.md
@@ -14,3 +14,5 @@ Checklist for community approval:
 - [ ] 2 week consultation with community
 - [ ] Community issues adequately addressed
 - [ ] News item about the release
+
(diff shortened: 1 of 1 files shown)


ljgarcia (commented):
@AlasdairGray do we have a document about how to move from draft to release? We do not have a tutorial but as a community review is involved, I think there is a document part of the governance, if yes, I suggest adding a link to this template from it... otherwise, how are people moving a profile from draft to release going to find it when created a release PR?
The idea is making the content of this PR findable.

AlasdairGray:
I've added a link to the appropriate section of the governance document, but I don't think that is what you were suggesting. The template should become discoverable when someone [creates a pull request](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates#pull-request-templates).

The template will only become visible once it is merged into the main branch.

ljgarcia (approved):
Looks good

PR #479, opened 2021-11-02
Fix link to crosswalk GSheet

AlasdairGray:


albangaignard (approved):
ok for the fixed link

PR #481, opened 2021-11-02
Add details of RDA18 and ELIXIR Data/InterOp meetings

AlasdairGray:


ljgarcia (approved):
Looks good to me.
For the future, maybe worth adding an optional registration link to the template.

PR #484, opened 2021-11-08
Embed BioHackathon reporting slides

AlasdairGray:


AlasdairGray:
@ljgarcia I've changed this to embed a presentation only containing the Bioschemas reporting slides. Hopefully this alleviates your concerns about sharing the presentation link.  
(I think this solution is much cooler too)

ljgarcia (approved):
Looks good, video failed in Chrome but worked in Firefox, it happens sometimes

PR #489, opened 2021-11-19
Fix broken link

AlasdairGray:


bedroesb (approved):
I still wonder why locally it redirects and online not

ljgarcia (approved):
Looks good

AlasdairGray:
> I still wonder why locally it redirects and online not

Software versions? Is there any way we can set up a Docker instance with the exact infrastructure that GitHub use?

PR #491, opened 2021-11-19
Unify redirections

ljgarcia:
Profile redirections were not always the same, this is an attempt to unify them

### DIFF
--- a/pages/_profiles/Event/0.2-DRAFT-2019_06_14.html
+++ b/pages/_profiles/Event/0.2-DRAFT-2019_06_14.html
@@ -1,15 +1,15 @@
 ---
 redirect_from:
-- "/devSpecs/Beacon/specification/"
-- "/devSpecs/Beacon/specification"
-- "/devSpecs/Beacon/"
-- "/devSpecs/Beacon"
-- "/specifications/drafts/Beacon/"
-- "/specifications/drafts/Beacon"
-- "/specifications/Beacon/"
-- "/specifications/Beacon"
-- "/profiles/Beacon/"
-- "/profiles/Beacon"
+- "/devSpecs/Event/specification/"
+- "/devSpecs/Event/specification"
+- "/devSpecs/Event/"
+- "/devSpecs/Event"
+- "/specifications/drafts/Event/"
+- "/specifications/drafts/Event"
+- "/specifications/Event/"
+- "/specifications/Event"
+- "/profiles/Event/"
+- "/profiles/Event"
 
 name: Event
 
--- a/pages/_profiles/Event/0.2-DRAFT-2019_06_14.html
+++ b/pages/_profiles/Event/0.2-DRAFT-2019_06_14.html
@@ -1,10 +1,15 @@
 ---
 redirect_from:
-- "devSpecs/Event/specification"
-- "devSpecs/Event/specification/"
-- "/devSpecs/Event/"
-- "/specifications/drafts/Event"
-- "/profiles/Event/"
+- "/devSpecs/Beacon/specification/"
+- "/devSpecs/Beacon/specification"
+- "/devSpecs/Beacon/"
+- "/devSpecs/Beacon"
+- "/specifications/drafts/Beacon/"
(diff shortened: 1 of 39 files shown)


ljgarcia:
@bedroesb if you come up with an idea on how to add this to a template or so, it would be great!

bedroesb:
@ljgarcia haha I was looking into it on how to specify the redirect_from on a global level, but I did not find a solution. 

at this moment pages have the permalink without the backslash, an I tried to redirect from links with a backslash, but it appears that this redirect is not working on collection level, and can probably be taken out.

```yml
collections_dir: pages
collections:
  types:
    output: true
    permalink: /:collection/:path
    redirect_from:
      - "/:collection/:path/"
  meetings:
    output: true
    permalink: /:collection/:path
    redirect_from:
      - "/:collection/:path/"
  tutorials:
    output: true
    permalink: /:collection/:path
    redirect_from:
      - "/:collection/:path/"
```

I hope you scripted the redirects because this seemed like a lot of work :D

ljgarcia:
It was a lot of copy&paste.
Those profiles that have a draft after a release are slightly different from the others that only have drafts or have a release as latest option

bedroesb:
good job nevertheless!

pages/_profiles/Event/0.2-DRAFT-2019_06_14.html
```diff
@@ -1,10 +1,15 @@
 ---
 redirect_from:
-- "devSpecs/Event/specification"
-- "devSpecs/Event/specification/"
-- "/devSpecs/Event/"
-- "/specifications/drafts/Event"
-- "/profiles/Event/"
+- "/devSpecs/Beacon/specification/"
```
AlasdairGray, line 3: Should these be Event?

pages/_profiles/Person/0.2-DRAFT-2019_07_19.html
```diff
@@ -1,10 +1,15 @@
 ---
 redirect_from:
-- "devSpecs/Person/specification"
-- "devSpecs/Person/specification/"
-- "/devSpecs/Person/"
-- "/specifications/drafts/Person"
-- "/profiles/Person/"
+- "/devSpecs/Beacon/specification/"
```
AlasdairGray, line 3: Shouldn't these be Person?

pages/_profiles/Sample/0.2-RELEASE-2018_11_10.html
```diff
@@ -1,11 +1,15 @@
 ---
 redirect_from:
-- "devSpecs/Sample/specification"
-- "devSpecs/Sample/specification/"
-- "/devSpecs/Sample/"
-- "/specifications/drafts/Sample"
-- "/specifications/Sample"
-- "/profiles/Sample/"
+- "/devSpecs/Beacon/specification/"
```
AlasdairGray, line 3: Shouldn't these be Sample?

AlasdairGray (changes requested):
I think there are few copy/paste errors

Really good job in making these consistent. I'd been slowly doing this as we had revisions of profiles.

pages/_profiles/Event/0.2-DRAFT-2019_06_14.html
ljgarcia: Yes, fixed in latest commit

pages/_profiles/Person/0.2-DRAFT-2019_07_19.html
ljgarcia: Yes, fixed in latest commit

pages/_profiles/Sample/0.2-RELEASE-2018_11_10.html
ljgarcia: Yes, fixed in latest commit

ljgarcia:
Three profiles had the incorrect name in the redirection (Event, Person, Sample), now fixed.

AlasdairGray (approved):

PR #494, opened 2021-11-23
Some fixes to the meeting pages

bedroesb:
I also set by default that iframes are responsive, have a shadow and rounded corners for consistency . I know it looks a bit silly since it takes the whole width, but I did not yet figure out how you can otherwise make iframes easily responsive so they don't break trough the window when resizing.


bedroesb:
I also experimented with a less wide website: https://bedroesb.github.io/bioschemas.github.io/

@AlasdairGray This is more like how it used to be, and I kinda like it, what do you think?

bedroesb:
@AlasdairGray I found a nice solution using bootstrap wrappers: https://getbootstrap.com/docs/5.1/helpers/ratio/

I applied the fix to all Iframes and now they scale+ keep the ratio

AlasdairGray:
That looks better in that the slides now fill the iframe and the list of participants is below the iframe.

Is there a way to limit the maximum size of the iframe to say ~600 px?

bedroesb:
@AlasdairGray Sure, I added it to the wrapper. The wrapper is a good mobile friendly solution, we just have to remember to add it to the new iframes being added

AlasdairGray:
> @AlasdairGray Sure, I added it to the wrapper. The wrapper is a good mobile friendly solution, we just have to remember to add it to the new iframes being added

I'll undoubtably forget

AlasdairGray (approved):
Looks good

PR #498, opened 2021-12-13
Fixing live deploy link and usecase link in profiles

bedroesb:
This PR will also link towards the profile tab + open the corresponding profile in the live deploys page


bedroesb:
right now I do the linking towards the open profile using jekyll on the link level:

```
<a href="{{spec.live_deploy | append: '#nav-profile?row=' | append: spec.name}}">
```

Because this required the least amount of changes.

bedroesb:
Live preview here: https://bedroesb.github.io/bioschemas.github.io/

AlasdairGray (approved):
Works for the sample of profiles that I tried

PR #511, opened 2022-01-28
Moving posts to news directory

bedroesb:


AlasdairGray (changes requested):
News items are showing with the oldest not the newest first; both on the news page and the widget on the right of the home page.

bedroesb:
@AlasdairGray good catch! This should be fixed now

AlasdairGray (approved):

PR #516, opened 2022-02-15
Render DDE Generated Profile JSON on Website

AlasdairGray:
This branch renders the profiles edited through the [Data Discovery Engine](https://discovery.biothings.io/view/bioschemas/) onto the website.

Rendering Tasks
- [ ] Include `schema:schemaVersion` to JSON used on website
- [ ] Update GitHub task labels so that there is a corresponding label for each profile name, e.g. `type: Tool` needs to be renamed to `type: ComputationalTool`

Prior to rendering, the JSON files need to be manipulated so that Jekyll's primitive JSON handling can be overcome. The conversions needed are:
- [x] Change the filename so that it does not have any `.` in it
- [x] Replace the following sequences
    - [x] `@id` → `jsonld-id`
    - [x] `@graph` → `jsonld-graph`
    - [x] `@context` → `jsonld-context`
    - [x] `@type` → `jsonld-type`
    - [x] `$validation` → `jsonld-validation`
    - [x] `$schema` → `jsonld-schema`
    - [x] `$ref` → `jsonld-ref`
    - [x] `rdfs:comment` → `rdfs-comment`
    - [x] `rdfs:label` → `rdfs-label`
    - [x] `rdfs:subClassOf` → `rdfs-subClassOf`


AlasdairGray:
All the replacements within the JSON-LD file are automatically done by simply reading the JSON-LD file in using the [Python JSON library](https://docs.python.org/3/library/json.html) and then writing it straight out. No other processing is required.

AlasdairGray:
I believe that is the functionality of the conversion script in place. It reads in a DDE generated `.json` file from the [Specifications repo](https://github.com/BioSchemas/Specifications), replaces the json-ld and json-schema keys (that jekyll can't process), and then writes the resulting `.json` file to the `_data/schemas` directory.

What is left to do for the script:
- Use input arguments to control which profile is converted and test with more than [ComputationalTool v1.0](https://github.com/BioSchemas/specifications/blob/master/ComputationalTool/jsonld/ComputationalTool_v1.0-RELEASE.json)
- Move to the [DDE repo](https://github.com/BioSchemas/bioschemas-dde)
- Set up GitHub actions chaining with the DDE conversion script

The task for this repo is to decide how we display the new schemas. In this test branch you can try out the ComputationalTool. The test file renders to [localhost:4000/CompTool.html](localhost:4000/CompTool.html)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "_data/specifications"]
+	path = _data/specifications
+	url = https://github.com/BioSchemas/specifications.git
```
bedroesb, line 3: @AlasdairGray do we still need this? If we pull straight from the specifications repo using requests, I guess we don't need a local copy with versioning in this repo?

AlasdairGray: Yes the submodule definition can go as we can't process the json files without first manipulating them using the script I developed here.

AlasdairGray:
@bedroesb I've moved the conversion script into the bioschemas-dde repo and extended it to process over the [configuration file](https://github.com/BioSchemas/bioschemas-dde/blob/main/specifications_list.txt) that Ginger uses for getting the data into DDE.

I've successfully generated the Jekyllised versions of the JSON files and loaded them into the `_data/schemas` directory so that we can use this branch/PR to focus on the rendering of them.

AlasdairGray:
@bedroesb I've made a start at rendering the profile information from the JSON-Schema files. I've taken the opportunity to redesign the page slightly. It'll be interesting to see what you think.

I've been testing with  (see screenshots)
- [localhost:ComputationalTool v1.0-RELEASE](http://localhost:4000/profiles/ComputationalTool/1.0-RELEASE)
- [localhost:Course v0.9-DRAFT](http://localhost:4000/profiles/Course/0.9-DRAFT-2020_12_08.html)

Screenshot 2022-02-18 at 16 35 56

Screenshot 2022-02-18 at 16 36 42

Screenshot 2022-02-18 at 16 37 00

bedroesb:
@AlasdairGray this is awesome! You really put some time in it these days. I will have a look at it next week :)

AlasdairGray:
I’ve made some progress onto the definition table, it is not easy going in liquid/jekyll. 

Given that we are already processing the JSON-Schema file to convert it into a form that Jekyll can process, should we just convert it into a YAML file with the content that we need for rendering on the site?

This would greatly simplify the logic that is implemented in the website rendering and allow us to do it in a language with proper support for the types of operations we are trying to undertake.

@bedroesb thoughts?

bedroesb:
I was looking into and I agree, to easiest is to process it in a Liquid digestible way (YAML). I will be way cleaner than any hacky approach otherwise. If we don't need changes to how it is visualized currently (in the website that is now live), we can format it with the same structure it is now in the frontmatter of those pages

ljgarcia:

PR #519, opened 2022-03-01
update Dataset draft profile version 0.5

albangaignard:
keyword cardinality is now "MANY"


AlasdairGray (approved):
I added in a news item

PR #526, opened 2022-03-23
FAIR-checker news

albangaignard:
Announcement of the FAIR-Checker release


AlasdairGray (approved):
Corrected label and added a bit about needing to use the Inspect option.

PR #534, opened 2022-04-26
Redirect fix

bedroesb:


bedroesb:
Maybe hold on to this one, have to check this further

bedroesb:
I can't get it to work. The fact that locally it behavious totally different compared to deployed does not make it easier. On other websites I got it to work well, but here for some reason I always have to make  a trade-off in some way...

PR #539, opened 2022-05-13
Added BiGCaT teaching material as live deployment

egonw:


AlasdairGray (approved):
Looks good.

@egonw can we add details of this to the news item for the TrainingMaterial 1.0 release?

PR #549, opened 2022-06-07
Add video and slides

AlasdairGray:


bedroesb (approved):

AlasdairGray:
Thanks for fixing the embedded slides.

bedroesb:
problem is that [bootstrap](https://getbootstrap.com/docs/5.0/helpers/ratio/) works with a wrapper div, so we would need some java script to add those parent divs if we want to do this automatically. I like to use the `ratio` classes because they are responsive

AlasdairGray:
I've added that code to any other place that we have a slideshow without it. I've just asked you to review the PR.

PR #553, opened 2022-06-20
Software page proposal

bedroesb:
Moved to a yaml file so contributors don't need to write html but can write markdown.

- [x] Mechanics
- [x] Software yaml
- [x] Porting all software tools

[Watch preview here](https://bedroesb.github.io/bioschemas.github.io/developer/software)

Screenshot from 2022-06-20 18-21-46


bedroesb:
@AlasdairGray , as promised, a proposal

AlasdairGray:
That is along the lines of what I had in mind. 

When we have multiple tools in a category, would it make sense to use the approach of the live deploys page of having the names listed with the ability to click to expand for details?

The other thing we should consider is whether it is worth keeping deprecated tools on the page or not.

bedroesb:
@BioSchemas/website  Some help with filling in the [YAML](https://github.com/bedroesb/bioschemas.github.io/blob/software-tabs/_data/software.yaml) file would be handy :) Please see the other examples for the syntax. they will render here: https://bedroesb.github.io/bioschemas.github.io/developer/software

bedroesb:
@gtsueng thanks! looks good! @AlasdairGray I added a comment in the YAML file as discussed in the call yesterday. What do you think should be done before merging?

AlasdairGray:
I've added the Schema.org validator under a validator category. @bedroesb does this need a code change somewhere?

bedroesb:
There where syntax errors because you used brackets/colons as list value. You can fix this by using quotes. 

another possibility is to not use markdown in the yaml but separate the nam/url as key values pairs.

AlasdairGray:
Thanks for fixing my error.

AlasdairGray:
@bedroesb is it intentional that the value of the `url` attribute is not used?

for the Schema.org validator I was expecting there to be a link to the tool itself by using that value

bedroesb:
Good catch! Forgot about that one. Where do we put it?

AlasdairGray:
Either below the tool name or even make the tool name a hyperlink.

bedroesb:
@AlasdairGray I updated so the URL is in the title now with an icon.

I also saw that we were still using an old version of font awesome for the icons, so I updated it to the latest version. I checked for compatibility issues but it seems fine

AlasdairGray:
For completeness, I'm stating here the tools that we have not ported to the new page.
- GOWeb: https://github.com/Bioschemas/bioschema-goweb  
  No longer required with move to DDE
- Google Dataset Search: https://toolbox.google.com/datasetsearch  
  No longer available. Been replaced by validator.schema.org which is listed
- GOCrawlIt: https://github.com/ricardoaat/bioschemas-gocrawlit  
- Validata: https://github.com/HW-SWeL/Validata  
  First attempt at a validation engine. No longer maintained.
- BuzzBang: https://github.com/buzzbangorg/buzzbang-doc/wiki  
 First attempt at a crawl and search engine. No longer maintained.
- MarkupBuilder  
  Implementation never shared, only ever a prototype

AlasdairGray (approved):
Looks good and a big improvement over our previous page. Thanks @bedroesb

PR #556, opened 2022-06-27
Update training material profile version

ljgarcia:


albangaignard (approved):
works in my local copy

PR #560, opened 2022-07-25
Add bioinformatics.ca to live deploys

AlasdairGray:


_data/live_deployments.json
```diff
@@ -1594,6 +1594,25 @@
...
       ]
+    },
+    {
+      "name": "Bioinformatics.ca",
+      "description": "Bioinformatics.ca, home of the Canadian Bioinformatics Workshops and bioinformatics resources in Canada.",
+      "url": "https://bioinformatics.ca/",
+      "profiles":[
+        {
+          "profileName": "Course",
+          "conformsTo": "1.0-RELEASE",
```
ljgarcia, line 1605: Is using RELEASE 1.0 (which has not been released yet) intended? Should it not be 0.8 DRAFT? I see the logic in directly using the coming RELEASE, just want to make sure it is ok and does not bring unintended issues on website, validators and so.

AlasdairGray: I'm not anticipating much change when we go to the 1.0-RELEASE so was just saving some edits. But if you feel that we should be correct then we can change things.

ljgarcia: I do not anticipate much change either and directly going for RELEASE 1.0 makes things easier for those adding the markup as there will be no need to update the conformsTo. All looks good to me.

ljgarcia (approved):

PR #564, opened 2022-08-10
Update Training.md

ljgarcia:
* Adds a new group lead to the Training group (as agreed with the other two leads)
* Uses YAML property ```lead``` in addition to ```current-leads``` as a temporary solution for leads to be listed in the group profile pages (a more permanent solution would be desirable, see https://github.com/BioSchemas/specifications/issues/586)

### DIFF
--- a/_includes/profile-start.html
+++ b/_includes/profile-start.html
@@ -127,7 +127,7 @@
       <div id="leadTable">
          {% for g in group %}
          {% for leader in g.lead %}
-         {% assign lead-id = "/people/" | append: leader %}
+         {% assign lead-id = "/people/" | append: leader.person %}
          {% assign author = site.people | where:"id", lead-id | first %}
          {% if author.homepage or author.github_username or author.orcid %}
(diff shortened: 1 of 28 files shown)


AlasdairGray:
I've made a lot of changes to @ljgarcia solution here but have fixed https://github.com/BioSchemas/specifications/issues/586 in the process

Leads are now shown with their start date if known. Fixed the display of information in the groups, profiles, and types pages.

@albangaignard or @ivanmicetic it would be good to get another review of this.

AlasdairGray:
Hi @albangaignard @ivanmicetic could you please have a look at this. Thanks

albangaignard (approved):
Thanks, group leaders are well displayed in my working copy.

ivanmicetic:
Looks good but has a small issue in style. If the Group Leaders list has one or two items (see Dataset profile or Protein type compared to FormalParameter profile), no bottom margin is applied: Other team members text isn't separated from the group leaders.

ivanmicetic (approved):
Looks good

PR #565, opened 2022-08-16
Update Course and CourseInstance draft profiles

ljgarcia:


AlasdairGray (approved):
I've fixed some links and file names. Looks good to add to the site.

PR #574, opened 2022-09-06
Updated TeSS' entry

fbacall:
* Added `TrainingMaterial`, `Course` and `CourseInstance`. 
* Updated version + example URL for `Event`.


AlasdairGray (approved):
Thanks @fbacall.

I corrected a couple of things relating to the highlights.

PR #579, opened 2022-10-17
Update Publications page with new presentations

AlasdairGray:
The publications page was missing several of our recent presentations. This has added several of them.

the PR also changes the mechanism by which publications are added. They are now added by putting them in a csv file. Hopefully this is more straightforward.


ljgarcia (approved):
Looks fine

PR #590, opened 2022-10-24
Update next community call date and time

AlasdairGray:
Update next call date and time. Time changed due to clocks changing in Europe.


ljgarcia (approved):
Looks good

PR #591, opened 2022-10-27
Centralizing contributors handling

bedroesb:
The code to render contributors was in total 3 times duplicated, so I made an include and changed from using lists to columns that act responsive. I also tried to make use of consistent naming of the attributes. This should also fix the spacing issue.

Preview can be found [here](https://bedroesb.github.io/bioschemas.github.io/)

Screenshot from 2022-10-27 09-57-25


AlasdairGray (approved):
Thanks for tidying this up. Looks good and appears to work well on the preview site.

PR #642, opened 2022-11-09
Add data dump details to the live deploy page

AlasdairGray:
Adds datafeed as a property of a site on the live deploy page and displays the address in the basic site information.

Added an extra tab that lists just the sites with datafeed deployments (see screenshot below).

Populated with sites that have a dataset download available.

Screenshot 2022-11-09 at 12 11 23


albangaignard (approved):
Works for me ! Thanks for this update !!

PR #687, opened 2022-11-11
Auto-Rendering Profile

sahar-frikha:


pages/_profiles/Dataset/0.6-DRAFT.html
```diff
@@ -0,0 +1,393 @@
...
+    be "Zung Scale" or "HAM-D" or "Beck Depression Inventory".
+
+
+    If there are several [variableMeasured](https://schema.org/variableMeasured) properties
+    recorded for some given data object, use a [PropertyValue](https://schema.org/PropertyValue)
+    for each [variableMeasured](https://schema.org/variableMeasured) and attach the
+    corresponding [measurementTechnique](https://schema.org/measurementTechnique).'
+  example: ''
+  expected_types:
+  - DefinedTerm
```
ljgarcia, line 199: DefinedTerm is not part of the range in schema.org, see https://schema.org/measurementTechnique
I know it would be useful and it is under discussion by the community but I do not think we have got to any agreement yet so I would wait to include it. @ivanmicetic @albangaignard @AlasdairGray any thoughts?

ljgarcia (changes requested):

_data/profile_versions.yaml
```diff
@@ -46,12 +46,12 @@ DataRecord:
   name: DataRecord
   status: deprecated
 Dataset:
-  latest_publication: null
+  latest_publication: 0.6-DRAFT
```
ljgarcia, line 49: DRAFT version should be 1.1 @gtsueng

ljgarcia (changes requested):

ljgarcia (approved):
Ready to go

PR #690, opened 2022-11-14
Correct to Vice Chair

AlasdairGray:


ljgarcia (approved):
Sorry I missed it before

PR #698, opened 2022-11-21
Auto-Rendering Profile

sahar-frikha:


ljgarcia:
@gtsueng @albangaignard @ivanmicetic there is a YAML error with the file _profiles/Gene/1.1-DRAFT.html I cannot figure it out. Could you please have a look? Thanks

ljgarcia (approved):
Issue with redirect for drafts vs releases but that has to be fixed for the whole website. Other than that, looks fine.

PR #719, opened 2022-12-21
Fixing group issues

bedroesb:
These changes will fix the problems described here: https://github.com/BioSchemas/specifications/issues/626

### DIFF
--- a/_includes/profile-start.html
+++ b/_includes/profile-start.html
@@ -136,9 +136,8 @@
          <tbody>
             <tr>
                <td>
-                  {% for g in group %}
+                  {%- assign g = site.groups | where:"identifier", page.group | first %}
                   <a href="{{g.url | relative_url}}">{{g.name}}</a>
-                  {% endfor %}
                </td>
                <td class="spec_links">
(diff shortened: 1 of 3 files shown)


bedroesb:
@ljgarcia I think this comment is not related to this PR? The fact that the url was not working had not to do with a redirect missing, but was a typo (quote was not closed)  that is already present for a very long time

bedroesb:
This PR is fixing problems that are there in the current deployment and does not hold back redirect changes

ljgarcia:
The comment is related to this PR. I understand the redirection problem does not come from this PR but I do not see the point in processing the PR right now knowing that the link to draft/release are wrong.

bedroesb:
I understand. But these wrong links towards release and draft are currently there already ... This is indeed a temporary solution if things will change, but one better than what now is in place. I don't see the reason of keeping broken urls online vs fixing them.

If the redirects change, we have to go through all urls in the website anyway. Just my opinion

ljgarcia (approved):

PR #724, opened 2022-12-22
Create BioChemEntity 0.8-DRAFT

gtsueng:
Note that this change reflects the jsonld generated by Alasdair, and addresses checkbox 1 of https://github.com/BioSchemas/specifications/issues/583

While the DDE allows for the generation of jsonld files, the auto-update to the website only applies to profiles and needs to eventually be updated to include types. For this reason, the html file was manually generated.

I will nest the `hasBioPolymerSequence` under `hasRepresentation` along with the other properties in version 0.9-DRAFT per 2nd checkbox of https://github.com/BioSchemas/specifications/issues/583

### DIFF
--- a/_data/type_versions.yaml
+++ b/_data/type_versions.yaml
@@ -2,7 +2,7 @@
 
 BioChemEntity:
     name: "BioChemEntity"
-    latest_publication:
+    latest_publication: "0.8-DRAFT"
     latest_release: "0.7-RELEASE-2019_06_19"
     status: "active"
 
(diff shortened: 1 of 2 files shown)


ljgarcia (changes requested):
@gtsueng could you please remove the redirection from 0.7-RELEASE-2019_06_19.html to avoid any possible conflict? Other than that, the PR is ok. Thanks

gtsueng:
@ljgarcia I have removed the redirects (only the redirect involving the term 'draft' was kept).

ljgarcia (approved):
Looks fine

PR #725, opened 2022-12-23
Double H1 titles

bedroesb:
It is recommended to only have one top level title, and since the title in the metadata of the md file is by default the H1 title, I would take away the one in the content.


bedroesb:
The problem:
Screenshot from 2022-12-23 14-38-06

We might want to think which is the best top title, because now there was a slight difference between them.

ljgarcia:
@gtsueng this PR looks good to me, please review and if it is also good for you, please accept, merge and delete branch. Thanks

PR #727, opened 2023-01-12
Auto-Rendering Profile

sahar-frikha:


gtsueng:
This was auto-generated as a result of merging [pull request #628](https://github.com/BioSchemas/specifications/pull/628) (Added subjectOf to MolecularEntity profile as optional property)

ljgarcia:
@gtsueng I want to close this PR without merging. Something went wrong with the 06-DRAFT so rather than continuing with the error, I suggest 07-DRAFT to fix the error and add subjectOf as optional. Please let me know if you agree.

What is wrong with 06-DRAFT? name, inChi and smiles should be kept but as either optional or recommended (@egonw what would be the way to go?). At least one of them MUST be present. We do not have a way to model this case in the specification, it should have to be implemented via the validator, it would be a special case for validation @albangaignard 
Also, sameAs has been added to all other profiles as recommended but it was removed from MolecularEntity. @egonw could it be kept as recommended as it in all the other profiles?

Once I have the comments from @gtsueng and @egonw I will (i) close this PR and create task for new 07-DRAFT fixing errors, or (ii) accept this PR and create task for new 08-DRAFT fixing errors. I would prefer (i), I do not see the point in accepting a new DRAFT that is continuing the errors from the previous one. Please let me know your thoughts. Thanks

gtsueng:
@ljgarcia I suspect the issue is that the conversion script cannot handle this particular edge case (requiring at least one of several options). This was the first time I've had to use this kind of JSON Schema validation rules for a Bioschemas profile. Do we have a prior example of how this type of rule should be displayed? If not we'll need to figure out how it should be displayed on the website.

ljgarcia:
@egonw what would you propose? Moving the three properties to "recommended" with a note on each saying that at least one of the three should be "minimum" and not having any listed "minimum" property?
How to deal with that in validation is a different issue, I would say. @gtsueng @albangaignard something to discuss later for validators?

egonw:
Sorry, pretty tied up in various things I have to give priority. I will comment asap (I need to dig up my notes from the BioHackathon with the alignment... @AlasdairGray, do you happen to still have your paper writing?)

ljgarcia (approved):
Looks fine

PR #735, opened 2023-02-22
Software page with bioschemas markup

albangaignard:
A Bioschemas page annotated with Bioschemas. 

Unfortunately it does not seem possible to automate the generation of markup in jekyll page frontmatters, from JSON data ...

### DIFF
--- a/_data/software.yaml
+++ b/_data/software.yaml
@@ -64,7 +64,7 @@ tools:
       - Issues for the [DDE Schema Playground, Registry, and Viewer](https://discovery.biothings.io/schema-playground) can be submitted to its [public repository](https://github.com/biothings/discovery-app/issues) 
 
     applicationCategory: viewer
-    url: https://discovery.biothings.io/schema-playground
+    url: https://discovery.biothings.io/registry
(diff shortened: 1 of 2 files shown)


bedroesb:
@albangaignard I think this manual way of filling this in is not really sustainable. I thought schema.org could also be included in the content of the page (outside of the head). Maybe we should just write a new include or layout to read out the yaml file and serve the bioschemas?

albangaignard:
Thansk for your feedback @bedroesb. I just followed the way we propose to annotate GitHub pages in the tutorial. But you are right, for the Bioschemas website itself, It's not the best approach. I will have a look at generating markup outside the jekyll frontmatter.

bedroesb:
@ljgarcia before merging I really think you should have revisioned the way tools are being added to the tools page. Either you load current tools from the bioschemas markup in the page, or you generate it based on the software yaml file with tools. Now people have to maintain both  ...

ljgarcia:
@bedroesb I understand but the request was adding the markup. Finding a clever to do so will be a new issue. Sorry it was not optimized since the beginning. @albangaignard please have a look to @bedroesb comments and maybe have a chat with him to brainstorm options.

PR #736, opened 2023-02-23
Auto-Rendering Profile

sahar-frikha:


gtsueng:
This is the auto-generated profile for the ComputationalWorkflow version 1.1 DRAFT

gtsueng:
The html has references to a temporary namespace (`bh2022GH`) which I tried to delete manually, but that caused errors, so I put it back.

PR #737, opened 2023-02-28
Update call date and time (moving to summer time)

ljgarcia:
Next community call on 2023-03-27 at 15UTC=16BST=17CEST


ivanmicetic:
Ok for 2023-03-27 at 15 UTC = 16 BST = 17 CEST = 8 PDT

PR #738, opened 2023-03-07
Hide Page/SMV/BMUSE buttons if missing or invalid exampleURL

ivanmicetic:
`live_deployments.json` lists resources with profiles with missing or invalid (`"exampleURL": "http://"`) `exampleURL` property. This fix will skip the rendering of Page, SMV and BMUSE buttons under Live Deploys page.

This will however leave wrong `exampleURL`s under Live Deploys/By Profile (e.g. under MolecularEntity profile). But this is tricky to fix and will require a new issue.


ivanmicetic:
Closing since resolved with #739

PR #739, opened 2023-03-08
Use name when deployments are listed / profile in the live deployments

bedroesb:
+ little bug fix so data dump tab is not always at the bottom

### DIFF
--- a/pages/_developer/liveDeploys.html
+++ b/pages/_developer/liveDeploys.html
@@ -85,15 +85,19 @@ bmuseHoverText: "Retrieve using Bioschemas Scraping service"
                             {%- assign deployhit = deploy.profiles | where:"exampleURL", live.exampleURL | where:"profileName", live.profileName | where:"highlight", live.highlight | first %}
                             {%- if deployhit %}
                             <div class="w-100 py-2">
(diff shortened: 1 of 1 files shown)


ivanmicetic:
Looks good. One thing, the `exampleURL` could be missing or invalid so I would change the code into something like this:
```
<div class="w-100 py-2">
    {%- if live.exampleURL != nil and live.exampleURL != "http://" %}
    <a class="me-2" href="{{live.exampleURL}}">
        {{deploy.name}}
    </a>
    {%- else %}
    <span class="me-2">{{deploy.name}}</span>
    {%- endif %}
    <span class="badge bg-dark text-light me-2">v{{live.conformsTo}}</span>
    {%- if live.highlight != nil %}
    <small class="text-muted me-2">{{live.highlight}}</small>
    {%- endif %}
</div>
{%- if live.exampleURL != nil and live.exampleURL != "http://" %}
<div class="ps-1">
```

bedroesb:
True! I updated the script

ivanmicetic (approved):
I will accept this while we copy datafeed url to exampleURL for all resources that have only datafeeds

PR #748, opened 2023-04-06
Auto-Rendering Profile

sahar-frikha:

### DIFF
--- a/_data/profile_versions.yaml
+++ b/_data/profile_versions.yaml
@@ -25,7 +25,7 @@ ComputationalWorkflow:
   name: ComputationalWorkflow
   status: active
 Course:
-  latest_publication: null
+  latest_publication: 1.1-DRAFT
   latest_release: 1.0-RELEASE
   name: Course
   status: active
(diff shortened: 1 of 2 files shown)


ljgarcia (approved):

ljgarcia:
@gtsueng @albangaignard 
I got an error that I easily fix but could be an indicative of something on the DDE or Actions side
@gtsueng could you please first check if the error comes from DDE? If not, please ping @albangaignard 

I will accept this PR but will put on hold the other two for CourseInstance and Sample as they present the same issue.

```
  Liquid Exception: Invalid Date: '"202346T165753"' is not a valid datetime. in /home/ubuntu/workspace/bioschemas.github.io/_layouts/profiles.html
             ERROR: YOUR SITE COULD NOT BE BUILT:
                    ------------------------------------
                    Invalid Date: '"202346T165753"' is not a valid datetime.
                    ------------------------------------------------
      Jekyll 4.3.1   Please append `--trace` to the `serve` command
                     for any additional information or backtrace.
                    ------------------------------------------------
```

gtsueng:
To my knowledge, there are no dates inside the JSON-LD file that was generated by the DDE. The JSON file loads fine in the DDE and works fine in the downstream automated process to ingest it into the DDE. I think it is from the GitHub actions, @albangaignard

PR #753, opened 2023-04-27
Update live_deployments.json

ljgarcia:
Adding profile TrainingMaterials to SIB learning platform in the live deploys


nsjuty (approved):
looks good

PR #757, opened 2023-05-01
Update taxon type

gtsueng:
The auto-profile rendering script does not work on types, so the pages for Taxon 1.0-RELEASE and TaxonName 1.0-RELEASE had to be manually generated

### DIFF
--- a/_data/type_versions.yaml
+++ b/_data/type_versions.yaml
@@ -123,13 +123,13 @@ Study:
 
 Taxon:
     name: "Taxon"
-    latest_publication: "1.0-RELEASE"
+    latest_publication: "0.4-DRAFT"
     latest_release: "1.0-RELEASE"
     status: "active"
 
 TaxonName:
     name: "TaxonName"
-    latest_publication: "1.0-RELEASE"
+    latest_publication: "0.1-DRAFT"
     latest_release: "1.0-RELEASE"
     status: "active"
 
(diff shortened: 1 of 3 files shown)


gtsueng:
@ljgarcia this is because the type_version yaml file has only two options:
'latest publication' and 'latest release'

Since the latest publication is actually the latest release, I had them both as version 1.0-RELEASE; however, it looks like the property 'latest publication' is used to populate the 'lastest draft' field so it's misleading.

In any case, I've updated 'latest publication' to be the old drafts which seems to have solve the issue.

ljgarcia (changes requested):
The current type drafts 0.4 for Taxon and 0.1. for TaxonName are being promoted to RELEASE. The drafts should no longer be listed in the draft list. That latest draft will be pointed to from the released type.

ljgarcia:
@frmichel please give me the go ahead for Taxon type
image
None of the properties list DefinedTerm as a possible value for the range? Is this correct? The description for the taxonRank indicates that defined terms are the best way to go...

ljgarcia:
@frmichel please also double check TaxonName
image

frmichel:
Our goal was to not constrain the choice of values for taxon ranks. There exist multiple lists of terms and URIs, some are more specific than others, and restricting to one of them might be perceived as a pb for some people.
So I'd suggest to leave it as is, for both Taxon and TaxonName. 
If needed, in a further version we could add a DefinedTerm if the community believes this would be a good thing.

ljgarcia (approved):
Looks good

PR #769, opened 2023-05-30
fixed error in example for alternateName: wrong quotation mark

DavidFichtmueller:
I noticed that the example for alternateName contains a wrong quotation mark, causing issues with the syntax parser: 
quotation mark error in alternateName example

I hope this can be fixed like this, and does not require a new release, as this is an error in the example and not in the specification itself.


ljgarcia:
Hi @albangaignard @gtsueng do you know where this error is coming? If the PR is directly accepted, will it break something wrt the integration with DDE? Should the update be done in DDE and then triggered in Bioschemas via the GitHub actions?

ljgarcia:
@albangaignard @gtsueng any comments?

gtsueng:
@ljgarcia the DDE does not ingest or touch any of the examples, so there will be no effect to the DDE. Near as I can tell, this file was created via the old method (GoWeb) (and the corresponding JSON-LD was created based on it), so neither Sahar's script nor the DDE should touch it.

gtsueng (approved):

PR #774, opened 2023-06-21
Fairchecker button live deploys

albangaignard:
Replaced BMUSE button by calls to the FAIR-Checker API. Two buttons: one for metdata harvesting, and one for Bioschemas profile validation.


albangaignard:
Here is how it looks like : 
Capture d’écran 2023-06-21 à 17 50 54

ljgarcia (changes requested):
@albangaignard could you please run some more tests? It does not work with some live deploys, for instance:
* https://www.alliancegenome.org/gene/RGD:620474 (no dct:conformsTo, created with REACT, the API does not produce results, the website produces a graph but no Bioschemas validation)
* https://aip.fipapatients.org/ (no dct:conformsTo, neither API nor website results)
* https://www.alliancegenome.org/gene/MGI:2442292 (API does not produce results, website keeps loading when trying to create the graph)
* https://aip.fipapatients.org/ (no dct:conformsTo, neither API nor website results)
* https://bar.biocomp.unibo.it/bar3/ (no dct:conformsTo, API does not produce results, website does)
* https://zbmed.github.io/damalos/ (no dct:conformsTo, API, does not give results, website does)
* url=https://workflowhub.eu/workflows/18 works fine with API and website

What could be happening with those not working with the API but with the website?

I would also suggest to include in this PR a new section FAIRChecker in https://bioschemas.org/tutorials/howto/howto_check_deploy so people can get some tips on why the validation could not work for their page

Thanks

ljgarcia:
Hi @albangaignard any news on this PR? Thanks

ljgarcia:
Hi @albangaignard any news on this PR? Thanks

albangaignard:
Hi @ljgarcia, thanks for coming back to me. 

If we taking this example https://zbmed.github.io/damalos/, here are the output of the periodic checks (https://github.com/BioSchemas/bioschemas-validation) : 

|     | URL                                                                                                                            |   HTTP_status |   nb_triples |   has_conforms_to | info                                                           |
|----:|:-------------------------------------------------------------------------------------------------------------------------------|--------------:|-------------:|------------------:|:---------------------------------------------------------------|
|   9 | https://zbmed.github.io/damalos/                                                                                               |           200 |           75 |                 0 | nan 

This means that harvesting is ok, we can retrieve 75 RDF triples. 

Then, there is no `conformsTo` property. We chose to evaluate Bioschemas profiles only when a `conformsTo` is specified. 
It explains why there is no results for validation with the API. 

On the FAIR-Checker website, we have another strategy, we try to infer the profile based on `rdf:type` properties. I have ongoing work to identify the "closest" profile but it's not "production" ready. 

Is it clearer ?

albangaignard:
Here is the output from the FC website : 

Capture d’écran 2023-09-06 à 13 30 23

ljgarcia (approved):
Thanks for checking, looks good

PR #780, opened 2023-07-13
Embedded twitter feeds do not work anymore

bedroesb:


ljgarcia (commented):
It could have been a glitch. This is what I see right now
image
Could it be working but nothing to show?

bedroesb:
@ljgarcia no, this function of twitter is just disabled by Elon Musk ;) One needs to login from now on to see things. We better take it away

bedroesb:
ELIXIR Europe for example also took their Twitter feed away from their website

ljgarcia (approved):

PR #782, opened 2023-07-28
Amend missing insulin image

M-casado:
I was reading through your documentation and the image for events is duplicated at https://bioschemas.org/tutorials/what_why_bioschemas (author @ljgarcia), and the one for insulin is missing.

I would have created a GH issue for this, but I believe these are not enabled at https://github.com/BioSchemas/bioschemas.github.io, so I did it myself. 

I assumed the image ``insulin.png`` I found unused in the folder ``images`` is the one you meant to use at the beginning, but I'm not sure.


bedroesb (approved):
You are correct! well spotted

PR #783, opened 2023-08-16
Delete BMUSE from live deploys

bedroesb:
Since this service is not live and we don't want urls that don't work


ivanmicetic (approved):
Looks good

PR #790, opened 2023-10-02
add entry for e!DAL-PGP in the "live deploy" section

arendd:
After multiple conversations in the past months with @AlasdairGray and later with @ljgarcia and @sneumann I think all issues are solved and the repository can be listed in the live deploy section.

### DIFF
--- a/_data/live_deployments.json
+++ b/_data/live_deployments.json
@@ -1748,6 +1748,27 @@
           "highlight": "Over 15'000 genes from multiple organisms."
         }
       ]
+       },
+    {
+      "name": "e!DAL-PGP: Plant Genomics and Phenomics Research Data Repository",
+      "description": "The Leibniz Institute of Plant Genetics and Crop Plant Research (IPK) and German Plant Phenotyping Network (DPPN) has jointly initiated the Plant Genomics and Phenomics Research Data Repository (PGP) as infrastructure to comprehensively publish plant research data. This covers in particular cross-domain datasets that are not being published in central repositories because of its volume or unsupported data scope, like image collections from plant phenotyping and microscopy, unfinished genomes, genotyping data, visualizations of morphological plant models, data from mass spectrometry as well as software and documents.",
+      "url": "https://edal-pgp.ipk-gatersleben.de",
+      "nodes": ["DE"],
+      "profiles": [
+       {
+          "profileName": "Dataset",
+          "conformsTo": "1.0-RELEASE",
+          "exampleURL": "http://dx.doi.org/10.5447/ipk/2022/5",
+          "highlight": "over 250 pages with markup"
+        },
+        {
+          "profileName": "Taxon",
+          "conformsTo": "0.6-RELEASE",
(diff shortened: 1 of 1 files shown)


ljgarcia (changes requested):
Hi @arendd thanks for the PR.
I checked http://dx.doi.org/10.5447/ipk/2022/5 and did not see the triple stating what Dataset profile version you are using. It would look like
```
    "http://purl.org/dc/terms/conformsTo": {
        "@id": "https://bioschemas.org/profiles/Dataset/1.1-DRAFT",
        "@type": "CreativeWork"
    }
```
For the Taxon example, the `comformsTo` was indeed there.

Not needed to process this PR but for future harvesting of these pages. I suggest you contact @albangaignard to find out why https://fair-checker.france-bioinformatique.fr/inspect cannot process your pages.

arendd:
Thank you @ljgarcia for your effort. I corrected this. Should be fine now, please have a look. @albangaignard could you give me a hint what could be the reason why FAIRChecker inspect is not working as expected, thank you.

ljgarcia (approved):
It now includes the necessary elements to be listed as a live deploy

ljgarcia:
@arendd I am accepting this PR as it now includes the elements needed for live deploys and passes the schema.org validation. However, I am leaving here some comments that could improve the markup.

- Be consistent in the way that `conformsTo` is included. For the Dataset it is expanded while for Taxon it uses the namespace (which is included in the schema.org context and because of this is correctly parsed). Either way, it would be better to expand the property name to `"http://purl.org/dc/terms/conformsTo":{ <profile name here>}`. According to Dublin Core, `conformsTo` has `Standard` as expected range, which is not the same as `schema:CreativeWork` so (if it was me) maybe no need to actually specify the type but only expand the `conformsTo` to its full URL.
```
"http://purl.org/dc/terms/conformsTo": {
  "@type": "CreativeWork",
  "@id": "https://bioschemas.org/profiles/Dataset/1.0-RELEASE"
},
```
vs
```
"dct:conformsTo":"https://bioschemas.org/profiles/Taxon/0.6-RELEASE",
```

- Schema.org is probably not aware of the `dwc` namespace. It will ignore it. However, other validators might complain about it as it cannot be properly parsed. I suggest to expand it to its full URL `http://rs.tdwg.org/dwc/terms/vernacularName`

-  Double check the markup for Taxon, I (human being) see an extra `]` at the end that makes no sense to me but the validator did not complain
```
{
		       "@context":"http://schema.org",
		       "@type":"Taxon",
		       "@id":"https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=info&id=4565",
		       "dct:conformsTo":"https://bioschemas.org/profiles/Taxon/0.6-RELEASE",
		       "name":"Triticum vulgare",
		       "taxonRank":"species",
		       "parentTaxon":"Triticum",
		       "dwc:vernacularName":"wheat"	
					}
		 	]
```

PR #791, opened 2023-10-08
Remove Galaxy Hub from LiveDeploys

sneumann:
Hi, the Galaxy/Events pages lost their Bioschemas markup. This regression is tracked in https://github.com/galaxyproject/galaxy-hub/issues/2177 with no timeline for a fix. Let's leave this PR open for a while, and hope we can cancel it when fixed by the Galaxy team. 
Yours, Steffen

### DIFF
--- a/_data/live_deployments.json
+++ b/_data/live_deployments.json
@@ -433,17 +433,6 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


ljgarcia (approved):
Looks good

PR #802, opened 2023-12-14
Add pages for new draft type LabProtocol

ljgarcia:
Creates draft type page v0.4 with updated hierarchy to HowTo


albangaignard (approved):
Looks good in my local working copy

PR #807, opened 2024-01-15
Labprotocol 05

ljgarcia:
Add changes requested in https://github.com/BioSchemas/specifications/issues/667
Type already in specifications see https://github.com/BioSchemas/specifications/pull/668


ivanmicetic (approved):
It looks good to me, renders correctly

PR #809, opened 2024-01-20
New database using ChemicalSubstance

egonw:
Right now with 42 entries (out of 600+ registered identifiers)

### DIFF
--- a/_data/live_deployments.json
+++ b/_data/live_deployments.json
@@ -1135,6 +1135,12 @@
           "exampleURL": "https://nanocommons.github.io/specifications/jrc/"
         },
         {
+          "profileName": "ChemicalSubstance",
+          "conformsTo": "0.4",
+          "exampleURL": "https://nanocommons.github.io/erm-database/",
+          "highlight": "Database of nanomaterials with European Registry of Materials identifier found on the internet"
+        },
+        {
           "profileName": "DataCatalog",
           "conformsTo": "0.3",
           "exampleURL": "https://nanocommons.github.io/datasets/"


ljgarcia:
@egonw `ChemicalSubstance` is already listed as a profile used by Nanocommons and only one example is needed per profile per site. I can see on the code that you are using different URLs. Do you want to update the example and add a highlight?

ljgarcia:
@ivanmicetic @nsjuty @albangaignard in case @egonw want to add a second link rather than updating the current example URL for `ChemicalSubstance` profile in Nanocommons, this is how it would look on the website.
image

If @egonw wants to add a second link, I would be against it as we only need only need one example per website. Also, we do not need/want to list all pages for a profile under the same domain. Any thoughts?

egonw:
> one example is needed per profile per site

Ah, interesting point! yeah, if the point is just to show which profiles are used, but they are two different projects (yeah, I'm involved in both). For me, listing both of them reflect the seriousness of adoption. But that would only work if the website would be showing both. Is that the case?

egonw:
> @ivanmicetic @nsjuty @albangaignard in case @egonw want to add a second link rather than updating the current example URL for `ChemicalSubstance` profile in Nanocommons, this is how it would look on the website.

All, since Steffen and I were in the Bioschemas telcon today that was not today (along with 3, 4 other people), we discussed a few things around `ChemicalSubstance` and `MolecularEntity`. This PR came up and we discussed the purpose of the live deploys (for us, for funders, for ELIXIR admins writing performance reports, etc).

And then I realized something. In the above 'NanoCommons' does not refer to a website (`schema:WebSite`) but to a grant (`schema:Grant`). So, another option is to unfold the NanoCommons grant into the separate websites.

Would that be in line what everyone has in mind for LiveDeploys?

ljgarcia:
Hi @egonw If NanoCommons is an umbrella for different projects, some of them with Bioschemas, some of them not, I would say it does make sense to have an entry per project. Could you please modify the PR accordingly? thanks

egonw:
Will do!

ljgarcia:
Hi @egonw any news on this PR? Thanks

egonw:
> any news on this PR?

@ljgarcia, ah, thanks for the ping! I have applied the suggestion.

ljgarcia (approved):
Looks good, thanks for the changes.

PR #817, opened 2024-06-18
Update index.html

ljgarcia:
Date and text displayed on the main page (which was outdated).


ivanmicetic:
I would remove the part "(15:00 BST/16:00 CEST)" and keep only 14:00 UTC so it is valid for the whole year.

ljgarcia:
UTC should be enough but we get people joining one hour late sometimes because got confused so, just in case, I added it. Some other info needs to be updated on regular basis, so keeping an eye on this one does not take a significant extra effort.

PR #822, opened 2024-08-01
New people added to Sample/BioSample

selgebali:
Added new people pages to tag in biosamples/samples


gtsueng (changes requested):
@selgebali @nsjuty Do the leads need to be changed as well?

nsjuty:
I think it should be Sara. That ok with you @selgebali ?

selgebali:
I changed it! sorry I missed the leads! :)

gtsueng (approved):

PR #823, opened 2024-08-06
Update years of ELIXIR Scientific Programme to 2024-2028

PhilReedData:
Better to include dates of current programme which still includes Bioschemas. The URL is the same.


ljgarcia:
@nsjuty could you please check the updated information is correct? Thanks

nsjuty:
its correct

nsjuty (approved):

PR #826, opened 2024-11-06
Fix link to markup generator http-only on the Software page

PhilReedData:
The https version of the markup generator appears to work but it is broken (reported here: https://github.com/BioSchemas/BioschemasMarkupGenerator/issues/16). Use the http version for now.

### DIFF
--- a/pages/_developer/software.html
+++ b/pages/_developer/software.html
@@ -39,7 +39,7 @@ bioschemas:
           "@type": Person
 
     - "@type": SoftwareApplication
-      "@id": https://www.macs.hw.ac.uk/SWeL/BioschemasGenerator/
+      "@id": http://www.macs.hw.ac.uk/SWeL/BioschemasGenerator/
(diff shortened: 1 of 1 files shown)


ljgarcia:
This generator should be deprecated as it is no longer maintain and it does not use the latest schema.org version.

ljgarcia:
@PhilReedData this generator has now been marked as deprecated and the one from DDE has been added. Would you agree to close this PR without further action?

ljgarcia:
Hi @PhilReedData it makes sense adding DDE schema playground to the menu but as replacement for the generator, we need to make sure that it is link within DDE that corresponds to generation. Could you please update the PR? Thanks

PhilReedData:
Yes, happy to update this PR to update the menus. Since I know this will conflict with the PR about breadcrumbs, I would like to finish that first, then I'll merge main into this branch and make a cleaner change. I am likely to get lost otherwise.

PhilReedData:
I've made the menu change in https://github.com/BioSchemas/bioschemas.github.io/pull/845 instead, since that is where I made the other menu changes, all in one PR. When that is merged, I will close this PR without any action.

PR #829, opened 2025-02-03
Add backlog of presentations from BioHackEU

PhilReedData:
Covers [Issue #623](https://github.com/BioSchemas/specifications/issues/623) adding three old BioHackathon Europe reports (2020, 2021, 2022) and one new one (2024).

I think we can abandon the [previous commit](https://github.com/BioSchemas/bioschemas.github.io/commit/24db021870faf910e4f565e5315cb4f6feeb4517) which appears to be done when publications could be entered directly in the publications.html page. Instead, I have added rows to the presentations.csv data file. There are three older reports and I have added the most recent. 

I tested this on a local installation and it worked fine. 
image


egonw:
@PhilReedData, what about a news item too?

PhilReedData:
Good idea. See PR https://github.com/BioSchemas/bioschemas.github.io/pull/831

PR #837, opened 2025-02-18
Tool updates

gtsueng:
Added DDE Markup generator and DDE Validator
Indicated that BMUSE and Bioschemas Generator are deprecated (in name and description)


ljgarcia (approved):
Looks good to me

PR #841, opened 2025-02-19
Update script and files to include associatedDisease equivalence

ljgarcia:
The equivalence towards schema.org types and properties for Bioschemas types included in pending requires a list of the included terms. We were missing assosciatedDisease, now included.


PhilReedData (approved):
As far as I can understand, this is working for me locally. I see the new lines containing 'schema:associatedDisease' in the ttl and jsonld files.

albangaignard (commented):
works for me !

PR #842, opened 2025-02-19
Add redirection from terms (Bioschemas ns) to types

ljgarcia:
We need https://bioschemas.org/terms/ (Bioschemas canonical namespace) to resolve to https://bioschemas.org/types, where we include links and signposting to the JSON-LD and TTL files.

### DIFF
--- a/pages/_types/drafts/index.html
+++ b/pages/_types/drafts/index.html
@@ -1,0 +1,5 @@
+---
+redirect_to: /types#nav-draft
+redirect_from:
+- "/devTypes/"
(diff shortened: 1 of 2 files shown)


ljgarcia:
We need this working for the integration with w3id.org

PhilReedData (approved):
It works locally for me. I visit the various drafts URLs and I land on the types page with the draft tab loaded. Very nice.

albangaignard (approved):
it also works for me, thanks !

PR #843, opened 2025-02-21
Create property pages

ljgarcia:
- Uses notebook to create the pages
- Modifies template so it takes into account expected types linked to schema.org rather than Bioschemas


ljgarcia:
Let's go with this approach by now. There are better ways, more compatible with Jekyll that could be implemented later

PhilReedData (approved):
I've deployed this branch locally, the property pages are working for me.

albangaignard (approved):
Everything is fine on my local copy, thanks !

PR #836, opened 2025-02-18
T7b1 - cologne hackathon collected modifications (Nick and Helena)

nsjuty:
Checklist for community approval:
- [ ] Use case present
- [ ] Crosswalk correctly linked with
  - [ ]  ~6 minimal properties, 
  - [ ] Examples for each property at minimal level
  - [ ] Examples for properties at recommended level
  - [ ] Mapping to terms in equivalent ontologies
 - [ ] Relevant issues closed on issue tracker and noted in change history
 - [ ] Examples in specification repo
 - [ ] Live deploys listing resources with latest version
   - [ ] Example one
   - [ ] Example two
- [ ] Training materials added
- [ ] 2 week consultation with community
- [ ] Community issues adequately addressed
- [ ] News item about the release

_Checklist based on requirements from [Bioschemas Governance](https://github.com/Bioschemas/governance/blob/master/governance.md#releasing-a-profile)._

PR #845, opened 2025-04-17
T7a6 Breadcrumbs

PhilReedData:
As raised in [issue 695](https://github.com/BioSchemas/specifications/issues/695), Cologne Hackathon task T7a6 is to add breadcrumbs across the Bioschemas website. We were advised to use Bootstrap (version 5.1 is already used) and to add page frontmatter metadata to determine the locations rather than rely on directories/paths.

I have implemented breadcrumbs on most of the site, with a few exceptions and one workaround (described below). The homepage is unchanged. I have also fixed a few menu items which were put into the wrong order in a previous PR #836. 

### Fly-out submenu workaround
I've also considered task T7a4: Fly-out submenu navigation, [issue 697](https://github.com/BioSchemas/specifications/issues/697). There was no flyout menu directly in the Bootstrap code, it would require much work with Javascript to borrow from elsewhere then thoroughly test. So I have instead put the nested menu items into one level deeper `ul` which means they display with a bullet. I believe this is clearer. I'm looking for feedback on this too.

### Breadcrumbs implemented fully
* `/tutorials/` Getting Started (22 pages)
* `/developer/` Deploy & Develop (2 pages)
* `/community/` Community (5 pages)
* `/groups/` Community > Groups (26 pages)
* `/about/` About (4 pages)
* Several index pages mentioned later.

These pages have additional front-matter page metadata to describe which breadcrumb to show. They use the default layout which calls the new `breadcrumb.html` include file. The new `<nav>` item uses the same design as the side menu boxes (tea green with rounded corners and shadow).

### Breadcrumbs working fine, displays slightly off
* `/profiles/` Specifications > Profiles 
* `/news/` Community > News
* `/meetings/` Community > Meetings
* `/useCases/` Community > Use Cases

These pages have a non-default layout (except their index pages). There are too many pages to add the front-matter to each of them, and I am not touching the mechanics of the specifications. I included the same `breadcrumb.html` file as with the default layout pages, but since it is called from the layout pages _later_, there it is too late to set the page metadata, so there are parameters passed to the included file with the same fields. Unfortunately, it is only possible to call the include from within the `<main>` element, so there is an extra margin above the navigation which I cannot remove. Operationally, the links work fine, but it is a little jarring to see the space appear and disappear when navigating between affected and unaffected pages. 

The Profiles pages work, though they show the breadcrumb with the profile version number but not its name. This is because the page title metadata of a profile is just the version number, without the title of the profile. I believe this to be an error of the profile implementation which is highlighted by my new code. I'm not going to work around it. Perhaps we should raise this matter separately?

### Breadcrumbs not working 
* `/types/` Specifications > Types
* `/properties/` Specifications > Properties.

These pages do not define a layout at all (except their index pages which use default). They are part of the specifications code which I am not touching. So I cannot add breadcrumbs to these.

### Screenshots
image
image
image
image
image
image
image

image

### DIFF
--- a/_includes/breadcrumb.html
+++ b/_includes/breadcrumb.html
@@ -1,0 +1,52 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="d-none">
+    <symbol id="house-door-fill" viewBox="0 0 16 16">
+        <path d="M6.5 14.5v-3.505c0-.245.25-.495.5-.495h2c.25 0 .5.25.5.5v3.5a.5.5 0 0 0 .5.5h4a.5.5 0 0 0 .5-.5v-7a.5.5 0 0 0-.146-.354L13 5.793V2.5a.5.5 0 0 0-.5-.5h-1a.5.5 0 0 0-.5.5v1.293L8.354 1.146a.5.5 0 0 0-.708 0l-6 6A.5.5 0 0 0 1.5 7.5v7a.5.5 0 0 0 .5.5h4a.5.5 0 0 0 .5-.5z"/>
+    </symbol>
+</svg>
+
+<div class="container my-1">
+    <nav aria-label="breadcrumb">
+        <ol class="breadcrumb p-3 bg-body-tertiary bg-light shadow rounded-3">
+            <li class="breadcrumb-item">
+                <a class="link-body-emphasis fw-semibold text-decoration-none" href="/">
+                    <svg class="bi" width="16" height="16" aria-hidden="true"><use xlink:href="#house-door-fill"></use></svg>
+                    Home
+                </a>
+            </li>
+            {% if page.breadcrumb.title1 %}
+            <li class="breadcrumb-item">
+                {% if page.breadcrumb.link1 %}
+                <a class="link-body-emphasis fw-semibold text-decoration-none" href="{{ page.breadcrumb.link1 | relative_url  }}">{{ page.breadcrumb.title1 }}</a>
+                {% else %}
+                {{ page.breadcrumb.title1 }}
+                {% endif %}
+            </li>
+            {% endif %}
+            {% if page.breadcrumb.title2 %}
+            <li class="breadcrumb-item">
+                {% if page.breadcrumb.link2 %}
+                <a class="link-body-emphasis fw-semibold text-decoration-none" href="{{ page.breadcrumb.link2 | relative_url  }}">{{ page.breadcrumb.title2 }}</a>
+                {% else %}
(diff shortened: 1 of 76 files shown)


PhilReedData:
I've improved the difference in non-default layout pages. Now, the breadcrumb is at the same height wherever it appears (using CSS transform). The width is still a little different on the non-default pages (described above). I think this is much less jarring now. 
image
image

ljgarcia:
@PhilReedData is this PR ready to go for revision? If yes, could you please remove the "draft" tag? thanks

ljgarcia:
This is how it looks for the page /profiles/ChemicalSubstance/0.4-RELEASE. Is the profile name missing?
image

PhilReedData:
The profile name is missing in the breadcrumb for all profiles. I used the page title, the same attribute used to populate the `<html><head><title>` field. All profile pages work this way, they show the version number in the html title but not the name of the profile. You normally see the html title in the browser tab/window. This issue is made more visible in the breadcrumbs.

Can we change the page title for the profiles to include the profile name before the version? That would be applied to the html head title and also to the breadcrumb. I don't know where that change would be required, presumably there is a script which generates the html for each profile?

PhilReedData:
... or should there be another level between, like
Home > Specifications > Profiles > ChemicalSubstance > 0.4 Release?

PhilReedData:
> @PhilReedData is this PR ready to go for revision? If yes, could you please remove the "draft" tag? thanks

It is nearly ready to go, I wanted to check about the profiles, types and properties pages first. It may be easier if I remove the Profiles breadcrumbs for consistency, then I could remove the draft tag?

PhilReedData:
I've removed the breadcrumbs from the Profiles pages since it isn't working. 
From the `layouts/profiles.html` page, I no longer add this first line after the frontmatter:
```
{% include breadcrumb.html title1="Specifications" title2="Profiles" link2="/profiles/" %}
```
Later (in a future PR), we can either: 
- Update the html head title include the name of the profile, or
- Write a more complicated layout script that makes the profile name appear as a separate crumb in the breadcrumb, eg Home > Specifications > Profiles > Beacons > 0.4 Release. (That is more work than I can do now.) 
  - Perhaps this solution could be extended to support Types and Properties too.

PhilReedData:
Discussed in https://github.com/BioSchemas/bioschemas.github.io/pull/826 I have implemented the agreed change as part of this PR instead. Replace the link in the navbar menu 'Markup Generator' to the DDE one, replacing the deprecated one.

PR #846, opened 2025-04-17
Move Ivan to former Steering Council members list

PhilReedData:
Update the Community page, Steering Council, to recognise that Ivan has stepped down.


albangaignard (approved):
everything is fine in my local copy, thanks !

PR #855, opened 2025-07-20
News item for ISMB/ECCB 2025

PhilReedData:
Let's get the news about ISMB/ECCB on the Bioschemas website. We will be directing people to the site all week. 

I've written this, adapted and reduced from a bigger article on the eScience Lab website (which Stian approved). 

We can add the DOI to Nick's slides when it becomes available.


albangaignard (approved):
Thanks @PhilReedData ! It renders nicely on my working copy.

PR #898, opened 2026-06-10
Auto-Rendering Profile

albangaignard:
This pull request has been created by a github action, after adding/updating a profile in the specifications repo.


ljgarcia:
@albangaignard I do not see changes in the specifications folder corresponding to this action. Also, the profile is Computational Tool rather than Life Sciences tool. Do you know which change in specifications repo triggered this?

PR #910, opened 2026-08-05
enumeration to isControl

selgebali:
added an expected type of [schema.org/Enumeration](http://schema.org/Enumeration) to allow for additional flexibility to isControl.


nsjuty (approved):
looks good

PR #913, opened 2026-08-14
Auto-Rendering Profile

albangaignard:
This pull request has been created by a github action, after adding/updating a profile in the specifications repo.


ljgarcia:
@albangaignard is this a new change triggered by the actions? It used to be a manual change. Triggered profile changes also come with more files, the ones corresponding to the profile. This PR seems to be incomplete

